PR #23, opened 2024-05-14
chore: Simplify type generation

nzakas:
This simplifies the generation of type definitions.

1. Uses a `// @ts-self-types` comment to specify types in the rolled-up packages, eliminating the need for a separate step to add a triple-slash directive.
2. Eliminates all `tsconfig.cjs.json` configurations. It turns out we can just copy over `index.d.ts` to `index.d.cts` and TypeScript will be happy.


nzakas:
> in that case why do we have `type.d.ts` (this seems like exact replica of `type.ts`)

`tsc` generates this. Maybe it's not needed, but I don't see the harm in leaving it. No need to add another step to the process.

> 2\. One more thing that I noticed is I removed all tsconfig.json and built the package just with plain rollup and still can see `type.ts` in the esm and cjs of dist folder as index.js imports declarations. In that case should we actually remove the copy plugin that does the exact same thing?

I'm guessing that you didn't delete `dist` before doing this. If you delete `dist` and try `npm run build` without Rollup copying over `types.ts`, you'll get a `tsc` error.

package.json
```diff
@@ -9,7 +9,8 @@
     "lint": "eslint .",
     "lint:fix": "eslint --fix .",
     "fmt": "prettier --write .",
-    "fmt:check": "prettier --check ."
+    "fmt:check": "prettier --check .",
+    "test:jsr": "npm run test:jsr --workspaces --if-present"
```
aladdin-add, line 13: consider adding `test:jsr` to the ci?

packages/object-schema/package.json
```diff
@@ -23,8 +23,9 @@
     "test": "tests"
   },
   "scripts": {
-    "build:prepend-type-ref": "node ../../tools/prepend-type-ref.js dist/esm/index.js",
-    "build": "rollup -c && tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json && npm run build:prepend-type-ref",
+    "build:cts": "node -e \"fs.copyFileSync('dist/esm/index.d.ts', 'dist/cjs/index.d.cts')\"",
+    "build": "rollup -c && tsc -p tsconfig.esm.json && npm run build:cts",
```
aladdin-add, line 27:
```suggestion
    "build": "tsc -p tsconfig.esm.json && rollup -c && npm run build:cts",
```

moving `tsc` to the first, so we can just use `rollup-plugin-copy` - `build:cts` is no longer needed.

nzakas: That doesn't work. Rollup generates the file that `tsc` evaluates.

package.json
nzakas: Good idea :+1:

.github/workflows/ci.yml
```diff
@@ -51,3 +51,19 @@ jobs:
...
+
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: "lts/*"
+
+            - name: Install Packages
+              run: npm install
+
+            - name: Run --dry-run
+              run: npm run test:jsr
```
aladdin-add, line 69: It needs to run build. It's fine to add it to the above job - after `npm test`. :)

nzakas: Good catch!

aladdin-add (approved):
LGTM. Waiting a few days before merging.

PR #27, opened 2024-05-27
fix: ConfigObject type

nzakas:
This renames the `BaseConfigObject` interface to `ConfigObject` and allows for any number of other named properties. I think this better reflects the reality of the package.


mdjermanovic (approved):
LGTM, thanks!

PR #68, opened 2024-06-24
feat: Add eslint/core package

### LINKED ISSUE #62: Change Request: eslint/core as types-only package (to start)

- [ ] `@eslint/compat`
- [ ] `@eslint/config-array`
- [ ] `@eslint/object-schema`

With language plugins, there a bunch of new type definitions that would be helpful to share from the start (for `Language`, `SourceCode`, etc.).

Because `eslint/core` will be the eventual home of runtime-agnostic ESLint APIs, I think it makes sense to start by creating this package to publish the types that are necessary for language plugins, and then slowly add more APIs and types over time as the rewrite unfolds.

(An alternate approach would be to update `@types/eslint` with this information, but that would create a dependency on DefinitelyTyped for publishing, which I'm not in favor of. If we publish types in `eslint/core`, then `@types/eslint` can always import those to keep things in sync.)

### Participation

### Additional comments

_No response_

nzakas:
#### Prerequisites checklist

This package creates the `@eslint/core` package, which is the future home of the runtime-agnostic rewritten core of ESLint.

Right now it just exports types for language plugins.

- Created the packaged
- Updated the release-please script
- Created type definitions

#### Related Issues

fixes #62

I wasn't sure if we should be testing the type definitions somehow? I'd defer to someone with more TypeScript knowledge than I on how that might be done.


packages/core/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "src"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "build": "tsc",
```
mdjermanovic, line 17: I'm not sure what's the intent of the build step, because it only generates a `dist/esm/types.d.ts` file but we're not publishing that file.

packages/core/src/types.ts
```diff
@@ -0,0 +1,371 @@
...
+}
+
+//------------------------------------------------------------------------------
+// Languages
+//------------------------------------------------------------------------------
+
+/**
+ * Represents a plugin language.
+ */
+export interface Language {
```
mdjermanovic, line 78: JS language also has [`visitorKeys`](https://github.com/eslint/eslint/blob/e2b16e2b72606162dce3d804bc80186b6c5ec0f9/lib/languages/js/index.js#L65C5-L65C16) property. Is it intentionally omitted from this interface because ESLint doesn't use it?

packages/core/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Sponsors
+
+The following companies, organizations, and individuals support ESLint's ongoing maintenance and development. [Become a Sponsor](https://eslint.org/donate) to get your logo on our README and website.
+
+<!-- NOTE: This section is autogenerated. Do not manually edit.-->
+<!--sponsorsstart-->
+<h3>Platinum Sponsors</h3>
+<p><a href="https://automattic.com"><img src="https://images.opencollective.com/automattic/d0ef3e1/logo.png" alt="Automattic" height="undefined"></a></p><h3>Gold Sponsors</h3>
+<p><a href="#"><img src="https://images.opencollective.com/guest-bf377e88/avatar.png" alt="Eli Schleifer" height="96"></a> <a href="https://engineering.salesforce.com"><img src="https://images.opencollective.com/salesforce/ca8f997/logo.png" alt="Salesforce" height="96"></a> <a href="https://www.airbnb.com/"><img src="https://images.opencollective.com/airbnb/d327d66/logo.png" alt="Airbnb" height="96"></a></p><h3>Silver Sponsors</h3>
```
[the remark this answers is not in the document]
harish-sethuraman: height is undefined? 🤔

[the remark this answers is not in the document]
mdjermanovic: That's how the script works, probably because there's no predefined fixed height for Platinum sponsors:

https://github.com/eslint/rewrite/blob/85018903815908bcfd940bbda347f3f17663e4bd/tools/update-readme.js#L35-L39

I guess it should be added, or the script should be updated to omit the height attribute.

packages/core/package.json
nzakas: It's to match other packages, where you can run `npm run build` to validate the TypeScript.

packages/core/src/types.ts
nzakas: I was trying to sync between what we were actually doing, but I think ESLint should be updated to use `visitorKeys` on the language object instead of the `SourceCode` object. Then I can add `visitorKeys` here.

packages/core/src/types.ts
```diff
@@ -0,0 +1,371 @@
...
+	 * Indicates if the original source contained a byte-order marker.
+	 * ESLint strips the BOM from the `body`, but this info is needed
+	 * to correctly apply autofixing.
+	 */
+	bom: boolean;
+
+	/**
+	 * The body of the file to parse.
+	 */
+	body: string | ArrayBuffer;
```
mdjermanovic, line 171: In ESLint, this is [`string | Uint8Array`](https://github.com/eslint/eslint/blob/e2b16e2b72606162dce3d804bc80186b6c5ec0f9/lib/linter/vfile.js#L76)

packages/core/README.md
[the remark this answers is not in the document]
nzakas: Yeah, we should fix the script.

Also WOW! Airbnb is now a platinum sponsor????

packages/core/src/types.ts
mdjermanovic: > ESLint should be updated to use `visitorKeys` on the language object instead of the `SourceCode` object.

That would be problematic for our JS language implementation, since it supports custom parsers and different parsers return different visitor keys.

packages/core/README.md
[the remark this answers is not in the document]
mdjermanovic: That's the info received from OC few days ago: https://github.com/eslint/eslint.org/commit/1741c34fa63a5ba86d9ba8c52a3f9da2f87150c4

packages/core/src/types.ts
```diff
@@ -0,0 +1,371 @@
...
+}
+
+/**
+ * Represents a location coordinate inside the source.
+ */
+export interface Position {
+	line: number;
+	column: number;
+	offset: number;
+}
```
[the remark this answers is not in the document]
nzakas: Oh yeah. I think I probably copy-pasted that from the Momoa types accidentally.

packages/core/src/types.ts
```diff
@@ -0,0 +1,371 @@
...
+//------------------------------------------------------------------------------
+// ASTs
+//------------------------------------------------------------------------------
+
+/**
+ * Represents an AST node or token with location information.
+ */
+export interface SyntaxElement {
+	loc: SourceLocation;
+	range?: [number, number];
```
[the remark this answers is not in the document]
nzakas: Yes it should good catch (also likely a copy-paste error).

packages/core/src/types.ts
nzakas: Another good catch. :+1:

packages/core/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "keywords": [
+    "eslint",
+    "core"
+  ],
+  "author": "Nicholas C. Zakas",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/eslint/rewrite/issues"
+  },
+  "homepage": "https://github.com/eslint/rewrite#readme",
```
harish-sethuraman, line 33: should we probably core's homepage here?

packages/core/src/types.ts
```diff
@@ -0,0 +1,370 @@
...
+	 * and ESLint should continue on to create a SourceCode object and run rules;
+	 * if false, ESLint should just report the error(s) without doing anything
+	 * else.
+	 */
+	ok: false;
+
+	/**
+	 * Any parsing errors, whether fatal or not. (only when ok: false)
+	 */
+	errors?: Array<FileError>;
```
mdjermanovic, line 216:
```suggestion
	errors: Array<FileError>;
```

I think this should be mandatory with `ok:false`.

packages/core/package.json
nzakas: We haven't done that for any of the other packages. If we want to do that, we can do so in a separate PR.

packages/core/src/types.ts
nzakas: Fair point. I think the problem here is that the JS implementation will be the exception case in this regard. It's unlikely that other languages will allow swapping out of parsers to something that returns an unknown AST.

So probably what we should be doing is `sourceCode.visitorKeys ?? language.visitorKeys` in ESLint, so that there's a way to override `visitorKeys` for a particular file if necessary, but it doesn't require every language to implement `visitorKeys` on `SourceCode`.

packages/core/src/types.ts
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * Represents a traversal step visiting the AST.
+ */
+export interface VisitTraversalStep {
+	type: "visit";
+	kind: 1;
+	target: SyntaxElement;
+	phase: 1 /* enter */ | 2 /* exit */;
+	args: Array<any>;
+}
```
mdjermanovic: There should be no `kind` property here?

https://github.com/eslint/eslint/blob/e2b16e2b72606162dce3d804bc80186b6c5ec0f9/lib/languages/js/source-code/source-code.js#L1227-L1243

packages/core/src/types.ts
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * Represents a traversal step calling a function.
+ */
+export interface CallTraversalStep {
+	type: "call";
+	kind: 2;
+	target: string;
+	phase?: string;
+	args: Array<any>;
+}
```
mdjermanovic: Same here, there's no `kind` on these objects.

https://github.com/eslint/eslint/blob/e2b16e2b72606162dce3d804bc80186b6c5ec0f9/lib/languages/js/source-code/source-code.js#L1261-L1265

Also, I'm not sure if we should specify `phase` here since ESLint is not using it.

packages/core/src/types.ts
mdjermanovic: Oh, I see now, `type` is translated into `kind` in the `TraversalStep` constructor.

packages/core/src/types.ts
mdjermanovic: And Linter indeed expects this property:

https://github.com/eslint/eslint/blob/7c78ad9d9f896354d557f24e2d37710cf79a27bf/lib/linter/linter.js#L939

packages/core/src/types.ts
```diff
@@ -0,0 +1,375 @@
...
+	/**
+	 * Applies configuration found inside of the source code. This method is only
+	 * called when ESLint is running with inline configuration allowed.
+	 */
+	applyInlineConfig?(): void;
+
+	/**
+	 * Applies language options passed in from the ESLint core.
+	 */
+	applyLanguageOptions?(): void;
```
mdjermanovic, line 273:
```suggestion
	applyLanguageOptions?(languageOptions: LanguageOptions): void;
```

packages/core/src/types.ts
```diff
@@ -0,0 +1,375 @@
...
+	getInlineConfigNodes?(): Array<SyntaxElement>;
+
+	/**
+	 * Applies configuration found inside of the source code. This method is only
+	 * called when ESLint is running with inline configuration allowed.
+	 */
+	applyInlineConfig?(): Array<{
+		configs: Array<InlineConfigElement>;
+		problems: Array<FileProblem>;
+	}>;
```
mdjermanovic, line 297:
```suggestion
	applyInlineConfig?(): {
		configs: Array<InlineConfigElement>;
		problems: Array<FileProblem>;
	};
```

This is just a single object, not an array.

https://github.com/eslint/eslint/blob/7c78ad9d9f896354d557f24e2d37710cf79a27bf/lib/languages/js/source-code/source-code.js#L1180-L1183

packages/core/src/types.ts
```diff
@@ -0,0 +1,375 @@
...
+	/**
+	 * Traversal of AST.
+	 */
+	traverse(): Iterable<TraversalStep>;
+
+	/**
+	 * Applies configuration found inside of the source code. This method is only
+	 * called when ESLint is running with inline configuration allowed.
+	 */
+	applyInlineConfig?(): void;
```
mdjermanovic, line 268: This seems like a duplicate, there's another `applyInlineConfig` below.

packages/core/src/types.ts
```diff
@@ -0,0 +1,375 @@
...
+	applyLanguageOptions?(): void;
+
+	/**
+	 * Return all of the inline areas where ESLint should be disabled/enabled
+	 * along with any problems found in evaluating the directives.
+	 */
+	getDisableDirectives?(): Array<{
+		directives: Array<Directive>;
+		problems: Array<FileProblem>;
+	}>;
```
mdjermanovic, line 282:
```suggestion
	getDisableDirectives?(): {
		directives: Array<Directive>;
		problems: Array<FileProblem>;
	};
```

This one also returns a single object, not an array.

https://github.com/eslint/eslint/blob/7c78ad9d9f896354d557f24e2d37710cf79a27bf/lib/languages/js/source-code/source-code.js#L1068-L1072

packages/core/LICENSE
JoshuaKGoldberg, line 1: [Docs] Following #52, it'd be good to have a decision doc for licensing. Even if it's just _"we need / want to use the same as existing ESLint licenses"_.

packages/core/README.md
```diff
@@ -0,0 +1,30 @@
+# ESLint Core
+
+## Overview
+
+This package is the future home of the rewritten, runtime-agnostic ESLint core.
+
+Right now, it simply exports the core types necessary to implement language plugins.
```
JoshuaKGoldberg, line 7: [Docs] As a language approachability point, nothing is "simple" / "simply".

```suggestion
Right now, it exports the core types necessary to implement language plugins.
```

packages/core/src/types.ts
JoshuaKGoldberg, line 1: Following #52 and #73: I don't feel able to review this file without understanding how this is meant to play with `@types/eslint` over time. Is it intentionally different? Have the tradeoffs been discussed somewhere?

packages/core/LICENSE
nzakas: There was no decision made. The OpenJS Foundation says any new projects must be licensed under Apache 2.0.

JoshuaKGoldberg: Is there a source that can be linked?

packages/core/src/types.ts
nzakas: We weren't consulted on the creation of `@types/eslint`, so I consider those to be non-canonical types. As we move forward, we'll be creating types the way we want and using those. Maybe `@types/eslint` will get smart and start importing our types to flesh out the package, but either way, I'm not going to worry too much about it.

packages/core/LICENSE
JoshuaKGoldberg: > Linked from where?

A decision doc, for visibility.

packages/core/src/types.ts
JoshuaKGoldberg: > non-canonical types

Practically speaking, those and typescript-eslint _are_ canonical types, as those are what the entire community uses at the moment. If the decision is to explicitly not attempt backwards compat, sure, that's a reasonable decision - but I don't think they can be ignored in decision-making summarily.

Issue #88, opened 2024-07-17
[Windows] Bug: npm ENOENT in `%APPDATA%/npm`


glenn2223:
### Which packages are affected?

- [ ] `@eslint/compat`
- [ ] `@eslint/config-array`
- [ ] `@eslint/core`
- [X] `@eslint/migrate-config`
- [ ] `@eslint/object-schema`

### Environment

Node version: 22.4.1
npm version: 10.8.1
ESLint version: 10.4.5 AND 11.0.0
Operating System: Windows 11 (Windows_NT 10.0.22631) plus my other local Windows 11 PC

### What did you do?

First, I hope this is the right place to open this issue. This is the only `npx` command I've had issue with recently so I assume it's nothing I've done wrong 😁

### What I did

Migrating my `.eslintrc.json` file using `npx @eslint/migrate-config ".eslintrc.json"`

### What did you expect to happen?

The command to run successfully

### What actually happened?

I have npm installed at `C:\Program Files\nodejs` and there's nothing in my AppData folder.

```diff
PS G:\MyPath> npx @eslint/migrate-config ".eslintrc.json"
npm error code ENOENT
npm error syscall lstat
npm error path C:\Users\MyUser\AppData\Roaming\npm
npm error errno -4058
npm error enoent ENOENT: no such file or directory, lstat 'C:\Users\MyUser\AppData\Roaming\npm'
npm error enoent This is related to npm not being able to find a file.
npm error enoent
npm error A complete log of this run can be found in: C:\Users\MyUser\AppData\Local\npm-cache\_logs\2024-07-17T12_50_57_438Z-debug-0.log
```

See workaround and details below 👇

### Link to Minimal Reproducible Example

N/A

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

Details taken from the log file

```diff
0 verbose cli C:\Program Files\nodejs\node.exe C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js
1 info using npm@10.8.1
2 info using node@v22.4.1
3 silly config load:file:C:\Program Files\nodejs\node_modules\npm\npmrc
4 silly config load:file:G:\MyPath\.npmrc
5 silly config load:file:C:\Users\MyUser\.npmrc
6 silly config load:file:C:\Users\MyUser\AppData\Roaming\npm\etc\npmrc
7 verbose title npm exec @eslint/migrate-config .eslintrc.json
[...]

PR #108, opened 2024-08-15
docs: Add workaround if npx fails

nzakas:
#### Prerequisites checklist

Adds instructions for what to do if the default `npx` command fails.

Added mention to manually install the package if the default `npx` command fails.

#### Related Issues

Refs https://github.com/eslint/rewrite/issues/88

### DIFF
--- a/packages/migrate-config/README.md
+++ b/packages/migrate-config/README.md
@@ -28,6 +28,13 @@ bunx @eslint/migrate-config .eslintrc.json
 
 The tool will automatically find your `.eslintignore` file in the same directory and migrate that into your new configuration file.
 
+If you receive any npm errors while running using `npx`, install the package manually first:
+
+```shell
+npm i @eslint/migrate-config
+npx @eslint-migrate-config .eslintrc.json
+```
+
 ### CommonJS Output
 
 By default, this tool generates an ESM file (`.mjs` extension). If you'd like to generate a CommonJS file instead, pass the `--commonjs` flag:
--- a/packages/migrate-config/README.md
+++ b/packages/migrate-config/README.md
@@ -32,7 +32,7 @@ If you receive any npm errors while running using `npx`, install the package man
 
 ```shell
 npm i @eslint/migrate-config
-npx @eslint-migrate-config .eslintrc.json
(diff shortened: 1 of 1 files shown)


harish-sethuraman (approved):
LGTM

packages/migrate-config/README.md
```diff
@@ -28,6 +28,13 @@ bunx @eslint/migrate-config .eslintrc.json
 
 The tool will automatically find your `.eslintignore` file in the same directory and migrate that into your new configuration file.
 
+If you receive any npm errors while running using `npx`, install the package manually first:
+
+```shell
+npm i @eslint/migrate-config
+npx @eslint-migrate-config .eslintrc.json
```
aladdin-add, line 35: made a few changes:
1. `-g`
2. add `@latest` to avoid caching issues.
```suggestion
npm i @eslint/migrate-config@latest -g
eslint-migrate-config .eslintrc.json
```

nzakas: I'm not sure we want to have people install globally that's why I'm still using `npx`.

aladdin-add, line 35: typo?
```suggestion
npx eslint-migrate-config .eslintrc.json
```

nzakas: I was trying to avoid `npx` globally installing, which is where the problems can occur. In this case, `npx` just runs locally.

nzakas: Actually, it should be `@eslint/migrate-config`

aladdin-add: Seems like both would work. `eslint-migrate-config` was defined: https://github.com/eslint/rewrite/blob/e0cfe185ec46fb4388663b6ff48d34f162859186/packages/migrate-config/package.json#L7

nzakas: Yes, both will work, I just think it's less confusing to stick to the same string everywhere.

aladdin-add: updated the suggestion. :)

aladdin-add (approved):
LGTM.

PR #110, opened 2024-08-22
feat: Add rule types

nzakas:
#### Prerequisites checklist

Add types for rules that will work regardless of the language being used.

I added several types that let us completely type check rules. These types are as generic as possible to allow for language plugins to override or better define their types, too.

These are *not* guaranteed to match the types in `@types/eslint`, which are very specific to the JavaScript language. The intent is to define something that will work and then see if we can backport into `@types/eslint` in a way that still works with the JavaScript rules.

Does the naming make sense for the types? I tried to be more specific to avoid ambiguity (i.e., `RuleFixer` is now `RuleTextEditor`).

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -58,7 +58,7 @@ export default [
 	// TypeScript
 	...tseslint.config({
 		files: ["**/*.ts"],
-		extends: [...tseslint.configs.strict],
+		extends: [...tseslint.configs.strict, ...tseslint.configs.stylistic],
 		rules: {
 			"no-use-before-define": "off",
 		},
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -58,7 +58,7 @@ export default [
 	// TypeScript
 	...tseslint.config({
 		files: ["**/*.ts"],
-		extends: [...tseslint.configs.strict, ...tseslint.configs.stylistic],
+		extends: [...tseslint.configs.strict],
 		rules: {
 			"no-use-before-define": "off",
 		},
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -269,7 +269,7 @@ export interface RuleContext<
 	 * The CommonJS path to the parser used while parsing this file.
 	 * @deprecated No longer used.
 	 */
-	parserPath: string;
+	parserPath: string | undefined;
 
 	/**
 	 * The rule ID.
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -37,14 +37,6 @@ export interface FileProblem {
 //------------------------------------------------------------------------------
 
 /**
- * Represents an AST node or token with location information in ESLint format.
- */
-export interface SyntaxElement {
-	loc: SourceLocation;
-	range: SourceRange;
-}
-
-/**
  * Represents the start and end coordinates of a node inside the source.
  */
 export interface SourceLocation {
@@ -281,13 +273,16 @@ export interface RuleContext {
 /**
  * Manager of text edits for a rule fix.
  */
-export interface RuleTextEditor {
+export interface RuleTextEditor<EditableSyntaxElement = unknown> {
 	/**
 	 * Inserts text after the specified node or token.
-	 * @param nodeOrToken The node or token to insert after.
+	 * @param syntaxElement The node or token to insert after.
 	 * @param text The edit to insert after the node or token.
 	 */
-	insertTextAfter(nodeOrToken: object, text: string): RuleTextEdit;
+	insertTextAfter(
+		syntaxElement: EditableSyntaxElement,
+		text: string,
+	): RuleTextEdit;
 
 	/**
 	 * Inserts text after the specified range.
@@ -298,10 +293,13 @@ export interface RuleTextEditor {
 
 	/**
 	 * Inserts text before the specified node or token.
-	 * @param nodeOrToken The node or token to insert before.
+	 * @param syntaxElement A syntax element with location information to insert before.
 	 * @param text The edit to insert before the node or token.
 	 */
-	insertTextBefore(nodeOrToken: object, text: string): RuleTextEdit;
+	insertTextBefore(
+		syntaxElement: EditableSyntaxElement,
+		text: string,
+	): RuleTextEdit;
 
 	/**
 	 * Inserts text before the specified range.
@@ -312,10 +310,10 @@ export interface RuleTextEditor {
 
 	/**
 	 * Removes the specified node or token.
-	 * @param nodeOrToken The node or token to remove.
+	 * @param syntaxElement A syntax element with location information to remove.
 	 * @returns The edit to remove the node or token.
 	 */
-	remove(nodeOrToken: object): RuleTextEdit;
+	remove(syntaxElement: EditableSyntaxElement): RuleTextEdit;
 
 	/**
 	 * Removes the specified range.
@@ -326,11 +324,14 @@ export interface RuleTextEditor {
 
 	/**
 	 * Replaces the specified node or token with the given text.
-	 * @param nodeOrToken The node or token to replace.
+	 * @param syntaxElement A syntax element with location information to replace.
 	 * @param text The text to replace the node or token with.
 	 * @returns The edit to replace the node or token.
 	 */
-	replaceText(nodeOrToken: object, text: string): RuleTextEdit;
+	replaceText(
+		syntaxElement: EditableSyntaxElement,
+		text: string,
+	): RuleTextEdit;
 
 	/**
 	 * Replaces the specified range with the given text.
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -167,11 +167,6 @@ export interface RulesMeta {
 	messages: Record<string, string>;
 
 	/**
-	 * The visitor keys for the rule.
-	 */
-	visitorKeys?: Record<string, string[]>;
-
-	/**
 	 * The deprecated rules for the rule.
 	 */
 	deprecated?: boolean | undefined;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -188,6 +188,7 @@ export interface RuleContext<
 	LangOptions = LanguageOptions,
 	Code extends SourceCode = SourceCode,
 	RuleOptions = unknown[],
+	Node = unknown,
 > {
 	/**
 	 * The current working directory for the session.
@@ -269,7 +270,7 @@ export interface RuleContext<
 	 * The report function that the rule should use to report problems.
 	 * @param violation The violation to report.
 	 */
-	report(violation: ViolationReport): void;
+	report(violation: ViolationReport<Node>): void;
 }
 
 // #region Rule Fixing
@@ -392,11 +393,11 @@ interface ViolationReportBase {
 }
 
 type ViolationMessage = { message: string } | { messageId: string };
-type ViolationLocation = { loc: SourceLocation } | { node: object };
+type ViolationLocation<Node> = { loc: SourceLocation } | { node: Node };
 
-export type ViolationReport = ViolationReportBase &
+export type ViolationReport<Node = unknown> = ViolationReportBase &
 	ViolationMessage &
-	ViolationLocation;
+	ViolationLocation<Node>;
 
 // #region Suggestions
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -185,8 +185,9 @@ export interface RulesMeta {
  * view into the outside world.
  */
 export interface RuleContext<
-	Code extends SourceCode = SourceCode,
 	LangOptions = LanguageOptions,
+	Code extends SourceCode = SourceCode,
+	RuleOptions = unknown[],
 > {
 	/**
 	 * The current working directory for the session.
@@ -262,7 +263,7 @@ export interface RuleContext<
 	/**
 	 * The rule's configured options.
 	 */
-	options: unknown[];
+	options: RuleOptions;
 
 	/**
 	 * The report function that the rule should use to report problems.
@@ -428,9 +429,10 @@ export type SuggestedEdit = SuggestedEditBase & SuggestionMessage;
  * The definition of an ESLint rule.
  */
 export interface RuleDefinition<
-	Visitor extends RuleVisitor = RuleVisitor,
-	Code extends SourceCode = SourceCode,
 	LangOptions = LanguageOptions,
+	Code extends SourceCode = SourceCode,
+	RuleOptions = unknown[],
+	Visitor extends RuleVisitor = RuleVisitor,
 > {
 	/**
 	 * The meta information for the rule.
@@ -442,7 +444,7 @@ export interface RuleDefinition<
 	 * @param context The rule context.
 	 * @returns The rule visitor.
 	 */
-	create(context: RuleContext<Code, LangOptions>): Visitor;
+	create(context: RuleContext<LangOptions, Code, RuleOptions>): Visitor;
 }
 
 //------------------------------------------------------------------------------
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -396,7 +396,7 @@ interface ViolationReportBase {
 	 */
 	fix?(
 		fixer: RuleTextEditor,
-	): RuleTextEdit | RuleTextEdit[] | IterableIterator<RuleTextEdit> | null;
+	): RuleTextEdit | Iterable<RuleTextEdit> | null;
 
 	/**
 	 * An array of suggested fixes for the problem. These fixes may change the
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -184,7 +184,10 @@ export interface RulesMeta {
  * information about the current state of the linting process and is the rule's
  * view into the outside world.
  */
-export interface RuleContext {
+export interface RuleContext<
+	Code extends SourceCode = SourceCode,
+	LangOptions = LanguageOptions,
+> {
 	/**
 	 * The current working directory for the session.
 	 */
@@ -221,13 +224,13 @@ export interface RuleContext {
 	/**
 	 * The source code object that the rule is running on.
 	 */
-	sourceCode: SourceCode;
+	sourceCode: Code;
 
 	/**
 	 * Returns the source code object that the rule is running on.
 	 * @deprecated Use `sourceCode` instead.
 	 */
-	getSourceCode(): SourceCode;
+	getSourceCode(): Code;
 
 	/**
 	 * Shared settings for the configuration.
@@ -243,7 +246,7 @@ export interface RuleContext {
 	/**
 	 * The language options for the configuration.
 	 */
-	languageOptions: LanguageOptions;
+	languageOptions: LangOptions;
 
 	/**
 	 * The CommonJS path to the parser used while parsing this file.
@@ -424,7 +427,11 @@ export type SuggestedEdit = SuggestedEditBase & SuggestionMessage;
 /**
  * The definition of an ESLint rule.
  */
-export interface RuleDefinition<T extends RuleVisitor = RuleVisitor> {
+export interface RuleDefinition<
+	Visitor extends RuleVisitor = RuleVisitor,
+	Code extends SourceCode = SourceCode,
+	LangOptions = LanguageOptions,
+> {
 	/**
 	 * The meta information for the rule.
 	 */
@@ -435,7 +442,7 @@ export interface RuleDefinition<T extends RuleVisitor = RuleVisitor> {
 	 * @param context The rule context.
 	 * @returns The rule visitor.
 	 */
-	create(context: RuleContext): T;
+	create(context: RuleContext<Code, LangOptions>): Visitor;
 }
 
 //------------------------------------------------------------------------------
@@ -498,7 +505,11 @@ export type RulesConfig = Record<string, RuleConfig>;
 /**
  * Represents a plugin language.
  */
-export interface Language {
+export interface Language<
+	LangOptions = LanguageOptions,
+	Code extends SourceCode = SourceCode,
+	RootNode = unknown,
+> {
 	/**
 	 * Indicates how ESLint should read the file.
 	 */
@@ -532,7 +543,7 @@ export interface Language {
 	/**
 	 * Validates languageOptions for this language.
 	 */
-	validateLanguageOptions(languageOptions: LanguageOptions): void;
+	validateLanguageOptions(languageOptions: LangOptions): void;
 
 	/**
 	 * Helper for esquery that allows languages to match nodes against
@@ -551,16 +562,19 @@ export interface Language {
 	 * throws errors for parsing errors but rather should return any parsing
 	 * errors as parse of the ParseResult object.
 	 */
-	parse(file: File, context: LanguageContext): ParseResult; // Future: | Promise<ParseResult>;
+	parse(
+		file: File,
+		context: LanguageContext<LangOptions>,
+	): ParseResult<RootNode>; // Future: | Promise<ParseResult>;
 
 	/**
 	 * Creates SourceCode object that ESLint uses to work with a file.
 	 */
 	createSourceCode(
 		file: File,
-		input: OkParseResult,
-		context: LanguageContext,
-	): SourceCode; // Future: | Promise<SourceCode>;
+		input: OkParseResult<RootNode>,
+		context: LanguageContext<LangOptions>,
+	): Code; // Future: | Promise<Code>;
 }
 
 /**
@@ -571,8 +585,8 @@ export type LanguageOptions = Record<string, unknown>;
 /**
  * The context object that is passed to the language plugin methods.
  */
-export interface LanguageContext {
-	languageOptions: LanguageOptions;
+export interface LanguageContext<LangOptions = LanguageOptions> {
+	languageOptions: LangOptions;
 }
 
 /**
@@ -607,7 +621,7 @@ export interface File {
 /**
  * Represents the successful result of parsing a file.
  */
-export interface OkParseResult<T extends object = object> {
+export interface OkParseResult<RootNode = unknown> {
 	/**
 	 * Indicates if the parse was successful. If true, the parse was successful
 	 * and ESLint should continue on to create a SourceCode object and run rules;
@@ -619,7 +633,7 @@ export interface OkParseResult<T extends object = object> {
 	/**
 	 * The abstract syntax tree created by the parser. (only when ok: true)
 	 */
-	ast: T;
+	ast: RootNode;
 
 	/**
 	 * Any additional data that the parser wants to provide.
@@ -652,8 +666,8 @@ export interface NotOkParseResult {
 	[key: string]: any;
 }
 
-export type ParseResult<T extends object = object> =
-	| OkParseResult<T>
+export type ParseResult<RootNode = unknown> =
+	| OkParseResult<RootNode>
 	| NotOkParseResult;
 
 /**
@@ -676,11 +690,16 @@ interface InlineConfigElement {
 /**
  * Represents the basic interface for a source code object.
  */
-interface SourceCodeBase {
+interface SourceCodeBase<
+	LangOptions = LanguageOptions,
+	RootNode = unknown,
+	SyntaxElementWithLoc = unknown,
+	ConfigNode = unknown,
+> {
 	/**
 	 * Root of the AST.
 	 */
-	ast: object;
+	ast: RootNode;
 
 	/**
 	 * The traversal path that tools should take when evaluating the AST.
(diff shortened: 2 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -75,6 +81,365 @@ export interface PositionWithOffset extends Position {
...
+ * A suggested edit for a rule violation.
+ */
+export type SuggestedEdit = SuggestedEditBase & SuggestionMessage;
+
+// #endregion
+
+/**
+ * The definition of an ESLint rule.
+ */
+export interface RuleDefinition<T extends RuleVisitor = RuleVisitor> {
```
nzakas, line 429: Using a type parameter here allows language plugins to insert their own `RuleVisitor` interface for better type checking. I imagine we can turn the one in `@types/eslint` into a `RuleVisitor` so those definitions will still work.

nzakas:
ping @eslint/eslint-tsc @JoshuaKGoldberg

eslint.config.js
```diff
@@ -58,7 +58,7 @@ export default [
 	// TypeScript
 	...tseslint.config({
 		files: ["**/*.ts"],
-		extends: [...tseslint.configs.strict, ...tseslint.configs.stylistic],
+		extends: [...tseslint.configs.strict],
```
JoshuaKGoldberg, line 61: 😄 not a fan of the stylistic config? Any reason in particular to remove this?

From running locally, I see two complaints:

```plaintext
/Users/josh/repos/rewrite/packages/core/src/types.ts
  108:8   error  A record is preferred over an index signature                                        @typescript-eslint/consistent-indexed-object-style
  389:12  error  Array type using 'Array<SuggestedEdit>' is forbidden. Use 'SuggestedEdit[]' instead  @typescript-eslint/array-type
```

* [`@typescript-eslint/array-type`](https://typescript-eslint.io/rules/array-type): purely stylistic for consistency. If you prefer `Array<...>` that's available as an option.
* [`@typescript-eslint/consistent-indexed-object-style`](https://typescript-eslint.io/rules/consistent-indexed-object-style): it's meant to be purely stylistic, but there are some changes between interfaces and records. If you want to use catch-all string signatures the way `RuleVisitor` is set up now, I'd just disable the rule with a comment. Either inline or in the config altogether.

packages/core/src/types.ts
```diff
@@ -75,6 +81,365 @@ export interface PositionWithOffset extends Position {
...
+/**
+ * Manager of text edits for a rule fix.
+ */
+export interface RuleTextEditor {
+	/**
+	 * Inserts text after the specified node or token.
+	 * @param nodeOrToken The node or token to insert after.
+	 * @param text The edit to insert after the node or token.
+	 */
+	insertTextAfter(nodeOrToken: object, text: string): RuleTextEdit;
```
JoshuaKGoldberg, line 293: The PRs in this repo are probably the heaviest users of `: object` I've seen this year 😄. I'm getting the impression types using them are split across two use cases:

* "Base" (_generalized_) types meant to be the constraints on type parameters, such as the current `RuleVisitor`
* _Generic_ types where there's a clear injectable, shared type

For "Base" (_generalized)_ types, it makes sense to use types like `object` and `unknown`. It's also common to name them something like `Any*` or `*Base` to make it clear that they're base types, not the ones end users directly use. Some of the types here are named like that, but some aren't - is that intentional?

For _generic_ types, it'd be more typical to use a type parameter. That way consumers of the type can substitute in their intended types, instead of having to do some `extends` strategy. Which would be very inconvenient for types like this with a lot of `: object`s!

Is there anything blocking making this type generic?

```suggestion
export interface RuleTextEditor<NodeOrToken> {
	/**
	 * Inserts text after the specified node or token.
	 * @param nodeOrToken The node or token to insert after.
	 * @param text The edit to insert after the node or token.
	 */
	insertTextAfter(nodeOrToken: NodeOrToken, text: string): RuleTextEdit;
```

Note that the same logic could be extended to any place that includes `: object` or `: unknown`.

* `ViolationLocation`'s `node`: which would make `ViolationReport` generic, which would make `RuleContext` generic.
* `RuleContext`'s `options` 
* `RuleVisitor`'s `node` and `parent` - though that type looks like it might be more of a "Base" _generalized_ type?

packages/core/src/types.ts
```diff
@@ -75,6 +81,365 @@ export interface PositionWithOffset extends Position {
...
+	/**
+	 * Indicates if the rule is generally recommended for all users.
+	 */
+	recommended?: boolean | undefined;
+}
+
+/**
+ * Meta information about a rule.
+ */
+export interface RulesMeta {
```
JoshuaKGoldberg, line 146: Similar to the note on `RuleTextEditor` below, `messages` and  could be made generic. We do this in typescript-eslint's types so that rule tests have a type error reported if an incorrect message ID is used.

For reference, https://github.com/typescript-eslint/typescript-eslint/blob/2ad3404690cd601d1bbb80daa7f861ab5bb127e4/packages/utils/src/eslint-utils/RuleCreator.ts#L32-L38:

```ts
export interface RuleWithMeta<
  Options extends readonly unknown[],
  MessageIds extends string,
  Docs = unknown,
> extends RuleCreateAndOptions<Options, MessageIds> {
  meta: RuleMetaData<MessageIds, Docs>;
}
```

That `Docs` generic is something to consider here too. Plugins can define their own docs, and we found often want to have their own properties on top of ESLint's. The `RuleMetadata` type for `docs` is an `&` intersection of the provided `Docs` and the ESLint standard `RuleMetaDataDocs`. https://typescript-eslint.io/blog/announcing-typescript-eslint-v8#custom-rule-metadocs-types

packages/core/src/types.ts
```diff
@@ -75,6 +81,365 @@ export interface PositionWithOffset extends Position {
...
+
+	/**
+	 * The messages that the rule can report.
+	 */
+	messages: Record<string, string>;
+
+	/**
+	 * The visitor keys for the rule.
+	 */
+	visitorKeys?: Record<string, string[]>;
```
JoshuaKGoldberg, line 170: Err, what is this? Rules will have their own visitor keys now?

packages/core/src/types.ts
```diff
@@ -75,6 +81,365 @@ export interface PositionWithOffset extends Position {
...
+	description?: string | undefined;
+
+	/**
+	 * The URL to the documentation for the rule.
+	 */
+	url?: string | undefined;
+
+	/**
+	 * The category the rule falls under.
+	 * @deprecated No longer used.
```
JoshuaKGoldberg, line 133: > `@deprecated`

A thought: it would be nice if these types don't need to include deprecated things... is there a strong need to include them?

For example: if these are meant to be used in ESLint core, but ESLint core wouldn't include them until the next major version, maybe anything meant to be dropped in that next major version can be omitted?

JoshuaKGoldberg (changes requested):
Sorry for the delay, and thanks for the ping! I still don't feel very confident reviewing functionally (https://github.com/eslint/rewrite/discussions/73), but have made some educated inferences just for what already exists in ESLint.

eslint.config.js
nzakas: Because I didn't think either of these warnings were helpful, and I didn't feel like fighting other rules that might come up (generally not a fan of applying rules blindly, especially ones that just seem to be enforcing stylistic preferences that I don't have).

packages/core/src/types.ts
nzakas: I don't have a strong opinion, it just seemed like it would be helpful to see something rather than not so that you know it wasn't an accidental omission in the type.

packages/core/src/types.ts
nzakas: I'll need to look at this a bit closer. The code snippet you pasted above makes my head hurt -- I have a hard time reasoning about what it actually means.

packages/core/src/types.ts
nzakas: In this case, there's only ever one instance of `RuleTextEditor`, which exists in the core and needs to work on everything...would that still be a case for a generic? 🤔 

(Still trying to wrap my brain around this.)

JoshuaKGoldberg: Type parameters are for when you want to substitute out different parts of a type for different uses. If there's only ever one instance of something, then you likely don't need type parameters - but then the type for `nodeOrToken` should be known as something more specific than `object`, right?

If the type for `nodeOrToken` is always going to be some known thing, then a different suggestion could be:

```suggestion
// or some existing type, or some union thereof...
export interface EditableNodeOrToken {
  /* ... */
}

export interface RuleTextEditor {
	/**
	 * Inserts text after the specified node or token.
	 * @param nodeOrToken The node or token to insert after.
	 * @param text The edit to insert after the node or token.
	 */
	insertTextAfter(nodeOrToken: EditableNodeOrToken, text: string): RuleTextEdit;
```

packages/core/src/types.ts
JoshuaKGoldberg: Yeah we're not huge fans of it ourselves 😄. A general summary if it helps...

What's happening is that each `RuleWithMeta` has _three_ "generic" parts that can be substituted in:

* What `Options` the rule has specified, which _must be_**\*** a `readonly` array 
* The `MessageIds` it can report, which _must be_**\*** a `string`
* The shape of `Doc`umentation, which doesn't need to be explicitly specified - it defaults to `unknown`

**\***_"must be"_ here is "assignable to": as in, whatever type we provide must satisfy that type:
* `readonly unknown[]` can be satisfied by, say, `[]` or `[{ allowInterfaces: boolean; allowObjects: boolean; }]`
* `MessageIds` can be satisfied by, say, `"noEmptyInterface" | `"noEmptyObject"``

It then _`extends`_ the interface `RuleCreateAndOptions`, which defines the standard JS `create()` method and typescript-eslint's `defaultOptions`. Plus then it adds a `meta` object of type `RuleMetaData`.

Putting it all together, this is a rough subset of [`@typescript-eslint/no-empty-object-type`](https://typescript-eslint.io/rules/no-empty-object-type):

```ts
type NoEmptyObjectType = RuleWithMeta<
  // Options
  [
    {
      allowInterfaces: boolean;
      allowObjects: boolean;
    }
  ],

  // MessageIds
  "noEmptyInterface" | "noEmptyObject",

  // Docs
  {
    recommended: 'recommended';
    requiresTypeChecking: true;
  }
>
```

packages/core/src/types.ts
```diff
@@ -75,6 +81,365 @@ export interface PositionWithOffset extends Position {
...
+ * times during the traversal.
+ */
+export interface RuleVisitor {
+	/**
+	 * Called for each node in the AST or at specific times during the traversal.
+	 */
+	[key: string]:
+		| ((node: unknown, parent?: unknown) => void)
+		| ((...unknown: unknown[]) => void);
+}
```
fasttime, line 115: How will this interface be used? Will it be extended to declare language-specific interfaces where `node` and `parent` have a particular type, or will rules simply create and return a `RuleVisitor` object?

packages/core/src/types.ts
nzakas: To fully flesh out what we have here: this same `RuleTextEditor` is used inside the core for applying fixes to any language that ESLint is linting (via a `Language` object). So for JavaScript, it works on ESTree(ish) nodes, while for Markdown it works on Unist nodes. While these two types of nodes only shared a `type`, there is no guarantee that any particular language will have that.

So, a rule is passing some representation of a node (or token) into these methods on `RuleTextEditor`. ESLint itself doesn't really need to know what those are, because we just pass them back into `Language#getRange()` to get the range information for that node or token. 

🤔 I suppose the audience for types here would be the rule developer, who may not be able to tell what to pass in. In that case, I could see it being helpful to specify the type as a parameter...but that would then bubble up into `RuleContext` and then up to `RuleDefinition`. Maybe that's not a bad thing in TypeScript land?

packages/core/src/types.ts
nzakas: It's designed to be extended but still work as a default. (see the `RuleDefinition` interface)

fasttime: I tried to declare an interface that extends `RuleVisitor` with something more specific, but it's not clear to me how the TypeScript syntax would look like ([**Playground link**](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgJQFcAbYANQQGcEZo4BvAKAEgB6AKnZbnbgGFMxUgBM46OsBwALOEgjC8yODCl4AggGUAKnDqZ4lMMGwJ0CbMoQBbYJTjDCUZAHNlq5VEwA3NJUEA6bnZWFgBtAGtgAE8ALjhKGCckZwBdGJZmAB84AAocuQU4wiRwuQB3JAAaODBMKGAUAH4ikvKkAEo4AF4APjgvCARhdozsvP8J4tKICpbpitCUzt7+weGAbkYAX0ZGGCijOABZKIA5eTwuhmUD4DiEpOd1-sFCO7gptrgtzcZkVAwODwJyIpAo1FoUDgoFQSGEdlB5CoNDoTGYEWi90SLjSuQK7xO5wU1Vq9SacUJF2WfQGQ02OyAA)).

fasttime: A possible solution: we could use generics in place of the actual argument types, e.g.:

```ts
export interface RuleVisitor<NodeType = unknown, ArgsType extends unknown[] = unknown[]> {
	[key: string]:
		| ((node: NodeType, parent?: NodeType) => void)
		| ((...unknown: ArgsType) => void);
}
```

Implementors could specify those types explicitly:

```ts
interface MyRuleVisitor extends RuleVisitor<MyNode> {
	[key: string]: (node: MyNode, parent?: MyNode) => void;
}
```

or

```ts
interface MyRuleVisitor extends RuleVisitor<never, [string, number, boolean]> {
	[key: string]: (arg1: string, arg2: number, arg3: boolean) => void;
}
```

[**Playground link**](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgJQFcAbYANQQGcEZoAeAOQgBNgAVATzDwF45CkAayQQA7kgA0cAIJQA5pU7c4oVEmaV+QkeIDaAXTh8BwsUgMA+OAG8AUAEgA9ACpnDuM7gBhTMVLM4dGgVHAALOBFWRCQ4GFC8aQBlNjhgzHhKbmwEdARsWIQAW2BNZkIoZDlY+NioTAA3NEpfADp3Z0cHXUFgDgAuOEoYCqQ5fT6HewAfOAAKWcjgAaZWJWApMEwoYBQAfmWWdi5gAEojK3qIBGYTyZn5lseTHSQB2QU1s54Lq5uAblsAF9bLYYMc4ABZDgrXg2WLHAZDEZyP5weq+QhLLSmcRwQEA2zIVAYHB4KFEUgUai0KAqEBqDQEEjkKg0ehQmFWOz2bq9RHDSrjOaLAYcw4bLY7GD7SHQw5fH7XAHAwkoNBYXCyiks6nQABMdIZmm1VLZUDoSGAjSgUl0SMqUiQhEKACM0FIXRAIKRMEh9Fyuj1+oMBaMhbMtnIAIz85FSSN6gZO13uuCRgDMA093uAvoVaN+ytsQA)

eslint.config.js
JoshuaKGoldberg: Understood. My (very obviously biased as typescript-eslint maintainer) recommendation would be to keep them enabled, but I understand why you'd want to avoid the stylistic preferences.

The terminology here is unfortunate: "stylistic" as used by that shareable config encompasses more than what "stylistic" means in ESLint core. It contains more than just consistency rules; it also has some real best practice violations and checks for likely bug causing syntax.

I'd recommend at least considering leaving the following rules (here or as a followup issue) that most folks see as definitive positives beyond just stylistic consistency:

* [`@typescript-eslint/adjacent-overload-signatures`](https://typescript-eslint.io/rules/adjacent-overload-signatures)
* [@typescript-eslint/consistent-type-definitions](https://typescript-eslint.io/rules/consistent-type-definitions)
* [@typescript-eslint/no-confusing-non-null-assertion](https://typescript-eslint.io/rules/no-confusing-non-null-assertion)
* [@typescript-eslint/no-inferrable-types](https://typescript-eslint.io/rules/no-inferrable-types)

https://typescript-eslint.io/rules/?=stylistic-xtypeInformation has the full listing of rules in `tseslint.configs.stylistic`.

JoshuaKGoldberg: Understood. My (very obviously biased as typescript-eslint maintainer) recommendation would be to keep them enabled, but I understand why you'd want to avoid the stylistic preferences.

The terminology here is unfortunate: "stylistic" as used by that shareable config encompasses more than what "stylistic" means in ESLint core. It contains more than just consistency rules; it also has some real best practice violations and checks for likely bug causing syntax.

I'd recommend at least considering leaving the following rules (here or as a followup issue) that most folks see as definitive positives beyond just stylistic consistency:

* [`@typescript-eslint/adjacent-overload-signatures`](https://typescript-eslint.io/rules/adjacent-overload-signatures)
* [`@typescript-eslint/consistent-type-definitions`](https://typescript-eslint.io/rules/consistent-type-definitions)
* [`@typescript-eslint/no-confusing-non-null-assertion`](https://typescript-eslint.io/rules/no-confusing-non-null-assertion)
* [`@typescript-eslint/no-inferrable-types`](https://typescript-eslint.io/rules/no-inferrable-types)

https://typescript-eslint.io/rules/?=stylistic-xtypeInformation has the full listing of rules in `tseslint.configs.stylistic`.

JoshuaKGoldberg: Understood. My (very obviously biased as typescript-eslint maintainer) recommendation would be to keep them enabled, but I understand why you'd want to avoid the stylistic preferences.

The terminology here is unfortunate: "stylistic" as used by that shareable config encompasses more than what "stylistic" means in ESLint core. It contains more than just consistency rules; it also has some real best practice violations and checks for likely bug causing syntax.

I'd recommend at least considering leaving the following rules (here or as a followup issue) that most folks see as definitive positives beyond just stylistic consistency:

* [`@typescript-eslint/adjacent-overload-signatures`](https://typescript-eslint.io/rules/adjacent-overload-signatures)
* [`@typescript-eslint/no-confusing-non-null-assertion`](https://typescript-eslint.io/rules/no-confusing-non-null-assertion)
* [`@typescript-eslint/no-inferrable-types`](https://typescript-eslint.io/rules/no-inferrable-types)

...and similarly at least considering these rules on _"just pick one and stick with it"_ consistency points that, in my experience, tend to get annoyingly split in a repo over time:

* [`@typescript-eslint/array-type`](https://typescript-eslint.io/rules/array-type)
* [`@typescript-eslint/consistent-type-definitions`](https://typescript-eslint.io/rules/consistent-type-definitions)

https://typescript-eslint.io/rules/?=stylistic-xtypeInformation has the full listing of rules in `tseslint.configs.stylistic`.

packages/core/src/types.ts
JoshuaKGoldberg: 👍 if the `RuleContext` cares about the type of nodes, and `RuleDefinition` therefore does as well, then it makes sense that `RuleDefinition` would have a type parameter for the kind of nodes.

In other words: if some consumer of `RuleContext`/`RuleDefinition` wanted to write a function that takes one of those in...

```js
// js version
function workOnContextNode(context: RuleContext) {
 context.sourceCode.ast.exampleProperty; // (or some equivalent)
}
```

...then TS users will want to be able to have that be well-typed without an `as`.


```js
interface MyLanguageRootAST {
  exampleProperty: true;
}

function workOnContextNode(context: RuleContext<MyLanguage>) {
 context.sourceCode.ast.exampleProperty;
}
```

packages/core/src/types.ts
snitin315: >maybe anything meant to be dropped in that next major version can be omitted?

I feel deprecated types should be present, otherwise,  users using deprecated properties may run into errors. We can remove the types when removing the deprecated properties in the next major.

packages/core/src/types.ts
nzakas: I think so. Going to dig into this now a bit more.

nzakas:
I've updated the definitions for `Language`, `SourceCode`, `RuleVisitor`, `RuleTextEditor`, and friends to be generic. @JoshuaKGoldberg would appreciate your feedback on this new attempt.

eslint.config.js
nzakas: Okay, I added it back.

packages/core/src/types.ts
```diff
@@ -75,6 +73,381 @@ export interface PositionWithOffset extends Position {
...
+export type RuleType = "problem" | "suggestion" | "layout";
+
+/**
+ * The type of fix the rule can provide:
+ * - `code` means the rule can fix syntax.
+ * - `whitespace` means the rule can fix spacing and indentation.
+ */
+export type RuleFixType = "code" | "whitespace";
+
+/* eslint-disable @typescript-eslint/consistent-indexed-object-style -- Needs to be interface so people can extend it. */
```
JoshuaKGoldberg, line 95: Aside: I hit this myself a few times recently. It's really annoying 😕. https://github.com/typescript-eslint/typescript-eslint/issues/7863 - added to my "todo soon" if nobody else gets to it...

packages/core/src/types.ts
```diff
@@ -309,11 +688,16 @@ interface InlineConfigElement {
 /**
  * Represents the basic interface for a source code object.
  */
-interface SourceCodeBase {
+interface SourceCodeBase<
+	LangOptions = LanguageOptions,
```
JoshuaKGoldberg, line 692: [Question] Can `LangOptions` be something that isn't assignable to `LanguageOptions`? Put another way: is it intentional that there's no `extends LanguageOptions` here?

packages/core/src/types.ts
```diff
@@ -75,6 +73,381 @@ export interface PositionWithOffset extends Position {
...
+	 * The meta information for the rule.
+	 */
+	meta?: RulesMeta;
+
+	/**
+	 * Creates the visitor that ESLint uses to apply the rule during traversal.
+	 * @param context The rule context.
+	 * @returns The rule visitor.
+	 */
+	create(context: RuleContext<LangOptions, Code, RuleOptions>): Visitor;
```
JoshuaKGoldberg, line 448: > `context: RuleContext<...>`

This isn't passing an argument for the `Node = unknown` type parameter. Is it intentional that `RuleDefinition` wouldn't know about the type of nodes in its language?

packages/core/src/types.ts
```diff
@@ -75,6 +73,381 @@ export interface PositionWithOffset extends Position {
...
+ * Represents the context object that is passed to a rule. This object contains
+ * information about the current state of the linting process and is the rule's
+ * view into the outside world.
+ */
+export interface RuleContext<
+	LangOptions = LanguageOptions,
+	Code extends SourceCode = SourceCode,
+	RuleOptions = unknown[],
+	Node = unknown,
+> {
```
JoshuaKGoldberg, line 192: > `LangOptions, Code, RuleOptions, Node`

👍 confirming that, as unweildy as having four type parameters feels, I think this is the right way to represent things. Each rule definition is specific to the _language_, _source code type_, _rule options_, and _language node_.

I'll note that a somewhat-equivalent-ish syntax to represent passing many type parameters would be to pass a single type parameter extending a shape. Roughly:

```suggestion
export interface RuleContextSettings {
	LangOptions: LanguageOptions;
	Code?: SourceCode;
	RuleOptions: unknown[];
	Node?: unknown;
}

/**
 * Represents the context object that is passed to a rule. This object contains
 * information about the current state of the linting process and is the rule's
 * view into the outside world.
 */
export interface RuleContext<Settings extends RuleContextSettings> {
```

...then later on referring to members of that type parameter:

```diff
- sourceCode: Code;
+ sourceCode: Settings["Code"];
```

It's conceptually like the refactor from a function with four parameter to a function with a single object parameter:

```diff
- function ruleContext(langOptions, code, ruleOptions, node) {
+ function ruleContext(settings) {

- langOptions;
+ settings.langOptions;
```

I've put up a comparison of what that could look like here: https://github.com/eslint/rewrite/compare/types-update-2...JoshuaKGoldberg:rewrite:types-update-settings-type-parameters.

Not saying either way is better, just showing another syntax as reference in case you prefer it.

JoshuaKGoldberg (dismissed):
OK! Thanks for the ping. I took a dive and think this is all looking right. I can tell you took care to work with the types, and it's showing - this feels very clean 🙂.

I left a few comments or questions, but nothing I can find that should block.

packages/core/src/types.ts
nzakas: `LanguageOptions` itself doesn't have any keys by default, so it seemed like `extends LanguageOptions` wasn't necessary? 

I was actually trying to reason through this choice, too, so any insights appreciated.

packages/core/src/types.ts
nzakas: Not intentional, just missed it. Thanks!

packages/core/src/types.ts
nzakas: Ah, interesting! I think I prefer the way it is, currently, but I consider myself educated. Thanks!

packages/core/src/types.ts
nzakas: Is there any value in having an empty interface? Isn't that just like putting `any` (or `unknown`)?

I was looking for something where everyone wouldn't need to create an index signature for their own rule visitor. For methods with esquery selectors, they'd need to override `[key: string]` to get a valid type for that kind of method.

packages/core/src/types.ts
nzakas: Actually, would that mean the type parameters could be specified in any order?

JoshuaKGoldberg: Yeah! And you're forced to use explicit names for them. `RuleContext<{ Key: Type, ... }>` instead of `RuleContext<Type, ...>`.

nzakas: Ooh nice! Okay, going to change to that.

packages/core/src/types.ts
fasttime: It's true, an empty interface can be assigned any value when the `strictNullChecks` TypeScript option is not set, so it would be only useful as an indicator for people reading the code.

Another option is keeping the index signature with a generic function type, for example:

```ts
export interface RuleVisitor {
  [key: string]: (...args: any) => void;
}
```

This would still not prevent implementors from adding methods that return non-void types, or from adding any other properties with arbitrary types.

nzakas: Went with this.

packages/core/src/types.ts
nzakas: I've switched all the types with way-too-may generic parameters to this model. The downside (as far as I can tell), is that TypeScript then either requires all of the properties or none of them -- the default is only used if nothing is passed, so I'm unable to assign defaults for each property. :thinking:

nzakas:
@fasttime @JoshuaKGoldberg can you take another look?

JoshuaKGoldberg (dismissed):
OK! I have no more inputs on the TypeScript types. Nicely done 🙁. This is a much more complex deep dive into TypeScript than most folks would want to get into at this stage of onboarding, I think, so it's great to see it so fleshed out.

A lot of the nuances (catch-all interfaces, `Options extends *Options`, ...) will probably have to evolve over time as the team figures out its preferred TypeScript practices. I'm excited to see how that goes, it'll be interesting!

packages/core/src/types.ts
```diff
@@ -75,6 +73,409 @@ export interface PositionWithOffset extends Position {
...
+/**
+ * An object containing visitor information for a rule. Each method is either the
+ * name of a node type or a selector, or is a method that will be called at specific
+ * times during the traversal.
+ */
+export interface RuleVisitor {
+	/**
+	 * Called for each node in the AST or at specific times during the traversal.
+	 */
+	[key: string]: (...args: unknown[]) => void;
```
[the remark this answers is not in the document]
nzakas: Ah interesting! I'll make that change.

packages/core/src/types.ts
```diff
@@ -75,6 +73,409 @@ export interface PositionWithOffset extends Position {
...
+
+	/**
+	 * The schema for the rule options. Required if the rule has options.
+	 */
+	schema?: JSONSchema4 | JSONSchema4[] | false | undefined;
+
+	/**
+	 * The messages that the rule can report.
+	 */
+	messages: Record<MessageIds, string>;
```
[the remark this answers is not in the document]
nzakas: Good catch!

packages/core/src/types.ts
```diff
@@ -75,6 +73,409 @@ export interface PositionWithOffset extends Position {
...
+	/**
+	 * The language options for the configuration.
+	 */
+	languageOptions: Options["LangOptions"];
+
+	/**
+	 * The CommonJS path to the parser used while parsing this file.
+	 * @deprecated No longer used.
+	 */
+	parserPath: string;
```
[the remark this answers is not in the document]
nzakas: Ah interesting. `parserPath` actually still exists on the `context` object, so I slightly favor leaving it here and marking as deprecated to make it clear. We can remove it from the type when we remove it from `context`.

packages/core/src/types.ts
```diff
@@ -75,6 +73,409 @@ export interface PositionWithOffset extends Position {
...
+	data?: Record<string, string> | undefined;
+
+	/**
+	 * The fix to be applied for the violation.
+	 * @param fixer The text editor to apply the fix.
+	 * @returns The fix(es) for the violation.
+	 */
+	fix?(
+		fixer: RuleTextEditor,
+	): RuleTextEdit | RuleTextEdit[] | IterableIterator<RuleTextEdit> | null;
```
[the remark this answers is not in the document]
nzakas: Nice!

packages/core/src/types.ts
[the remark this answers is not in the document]
fasttime: Makes sense :+1: I think the type should be `string | undefined` because `parserPath` is always undefined with flat config.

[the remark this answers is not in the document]
nzakas: Done

fasttime (approved):
LGTM, thanks!

PR #117, opened 2024-09-03
chore: fully type check packages/*/src files

JoshuaKGoldberg:
#### Prerequisites checklist

Creates a root-level `tsconfig.json` that can type check all project files, along with a `ci.yml` step to do so in CI.

This PR includes two common TypeScript practices for monorepos:

* A `tsconfig.base.json`: to unify & share the common TSConfig settings used by all projects
* A root-level `tsconfig.json`: so editors have that includes all files

That root-level `tsconfig.json` is necessary for typed linting with the recommended [typescript-eslint project service](https://typescript-eslint.io/packages/parser#projectservice), as noted by @snitin315 and myself as a followup in #90 -> https://github.com/eslint/rewrite/pull/90#discussion_r1702880458.

Also fixes a few type errors here and there. I'm posting comments in the PR.

Notably, this PR does _not_ set up [project references](https://www.typescriptlang.org/docs/handbook/project-references.html). Doing so requires touching files on disk, which I'm not confident enough in this repo to do on my own unprompted. Instead, the root-level `tsconfig.json` has `noEmit: true` so it's purely used for type checking.

#### Related Issues

Followup to #90, which is a PR. Would the team like me to file more granular issues? I wasn't sure how much the team wants some or all of this change. My intent is to continue enabling strict TypeScript flags after this, unless directed otherwise.

I'm applying the practices I see as common + good in TypeScript-land, but am not totally sure I interpreted the existing repo setup right. Very much seeking to understand. 🙂


.github/workflows/ci.yml
```diff
@@ -27,6 +27,9 @@ jobs:
             - name: Lint Files
               run: npm run lint
 
+            - name: Type Check Files
+              run: npm run tsc
```
JoshuaKGoldberg, line 31: [Question] Does the team want to type check in CI separate from the build step? As in, was it intended that `npm run build` should be what type-checks?

CONTRIBUTING.md
JoshuaKGoldberg, line 1: [Question] I ended up writing this because I wanted to add a mention of how to type-check, and didn't know where to put it... is this the right place for rewrite contributing docs?

package.json
```diff
@@ -27,6 +28,7 @@
     "!(*.{js,ts})": "prettier --write --ignore-unknown"
   },
   "devDependencies": {
+    "@types/mocha": "^10.0.7",
```
JoshuaKGoldberg, line 31: [Explanation] This isn't strictly necessary with `"noImplicitAny": false`. But it's just good to have types for dependencies in general.

packages/plugin-kit/src/config-comment-parser.js
```diff
@@ -9,7 +9,6 @@
 // Imports
 //------------------------------------------------------------------------------
 
-// @ts-ignore -- don't feel like fighting with TypeScript right now
```
JoshuaKGoldberg, line 12: [Question] I don't get any type errors in my editor or from `npm run build` on `main` after removing this comment directive. Is there another command I'm missing?

scripts/build.js
```diff
@@ -67,8 +67,8 @@ async function calculatePackageDependencies(packageDirs) {
 }
 
 /**
- * Creates an array of directories to be built in order to sastify dependencies.
- * @param {Map<string,{name:string,dir:string,dependencies:Set<string>}} dependencies The
+ * Creates an array of directories to be built in order to satisfy dependencies.
+ * @param {Map<string,{name:string,dir:string,dependencies:Set<string>}>} dependencies The
```
JoshuaKGoldberg, line 71: [Fix] Typo: missing `>` closing out `Map`.

tsconfig.base.json
```diff
@@ -0,0 +1,17 @@
...
+    "emitDeclarationOnly": true,
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "strict": true,
+    "target": "ES2022",
+
+    // TODO: Over time, enable these strict options
+    "noImplicitAny": false,
+    "strictNullChecks": false,
+    "useUnknownInCatchVariables": false
```
JoshuaKGoldberg, line 15: [Explanation] `strict` is generally a best practice, and has flags added to it over time. For anything that can't be added yet, the common practice is to disable them specifically with a TODO comment. Kind of like enabling a recommended preset in an ESLint config, with `rules` overrides.

I know there was mention of not wanting some or all of these - is here a good place to discuss?

types/levn.d.ts
JoshuaKGoldberg, line 1: [Question] I ended up adding as much as I could here. I plan on sending to DefinitelyTyped soon. Did I get the types right? (I've never used `levn`)

tsconfig.base.json
nzakas: I think strict vs. not should be opened up as a separate discussion.

.github/workflows/ci.yml
nzakas: `npm run build` is intended to do everything in preparation for a package being published (Including `tsc`, running additional scripts, running Rollup, etc.).

If we want to highlight type checking errors, then I think it's fine to have as a separate CI check.

CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Code of Conduct
+
+This project adheres to the [OpenJS Foundation Code of Conduct](https://eslint.org/conduct). We kindly request that you read over our code of conduct before contributing.
+
+## Commands
+
+### Building
+
+[Rollup](https://rollupjs.org) and [TypeScript](https://www.typescriptlang.org) are used to turn source files in `packages/*/src/` into outputs in `packages/*/lib/`.
```
nzakas, line 13:
```suggestion
[Rollup](https://rollupjs.org) and [TypeScript](https://www.typescriptlang.org) are used to turn source files in `packages/*/src/` into outputs in `packages/*/dist/`.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+
+```shell
+npm run build
+```
+
+### Linting
+
+ESLint is linted using ESLint.
+[Building](#building) the project must be done before it can lint itself.
```
nzakas, line 22: Did you find this to be true? Linting generally happens on source files, so not sure why building would be required first.

packages/plugin-kit/src/config-comment-parser.js
nzakas: Leftover comment from before I figured out how to get it to work. :smile:

types/levn.d.ts
nzakas: If you're planning on contributing to DefinitelyTyped, I'd rather wait for that than create a new top-level directory.

Also, I'm not familiar enough with levn to really know if the types make sense. I just hacked something together that made `tsc` happy and seemed to make sense based on the docs.

CONTRIBUTING.md
JoshuaKGoldberg: Yes, it has a dependency on its own files right now:

```plaintext
$ npm run lint

> eslint-rewrite@1.0.0 lint
> eslint .


Oops! Something went wrong! :(

ESLint: 9.9.1

Error: Cannot find module '/Users/josh/repos/rewrite/node_modules/@eslint/config-array/dist/cjs/index.cjs'
    at createEsmNotFoundErr (node:internal/modules/cjs/loader:1256:15)
    at finalizeEsmResolution (node:internal/modules/cjs/loader:1244:15)
    at resolveExports (node:internal/modules/cjs/loader:628:14)
    at Module._findPath (node:internal/modules/cjs/loader:718:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1205:27)
    at Module._load (node:internal/modules/cjs/loader:1045:27)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:215:24)
    at Module.require (node:internal/modules/cjs/loader:1304:12)
    at require (node:internal/modules/helpers:123:16)
```

Running a build fixes that:

```plaintext
$ npm run build

# ... build output

 $ npm run lint

> eslint-rewrite@1.0.0 lint
> eslint .

$ |
```

.github/workflows/ci.yml
JoshuaKGoldberg: Great. I've found it to be useful to have explicit checks for each of (lint, build, test, etc.) split out. What happens over time in larger repos is that type-checking time grows to be much much longer than build times, so folks reasonably want to be able to see results from tsc on its own separately from / in parallel to build. Both in CI and locally.

tsconfig.base.json
JoshuaKGoldberg: 👍 #118

types/levn.d.ts
JoshuaKGoldberg: 👍 https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70460

CONTRIBUTING.md
nzakas: I think it's fine for now. I'm always torn between `CONTRIBUTING.md` and `README.md`. I think it's more important that the information exists *somewhere* rather than worrying about exactly where at this point.

packages/plugin-kit/src/config-comment-parser.js
```diff
@@ -123,7 +122,8 @@ export class ConfigCommentParser {
 	parseJSONLikeConfig(string) {
 		// Parses a JSON-like comment by the same way as parsing CLI option.
 		try {
-			const items = levn.parse("Object", string) || {};
+			const items =
+				/** @type {RulesConfig} */ (levn.parse("Object", string)) || {};
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Great question: either works, but as-is is more precise.

From https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts, `@type` JSDoc annotations act as assertions[^1] on the parenthesized expression. Right now that expression includes only the `levn.parse(...)` call. Essentially it's `(levn.parse(...) as RulesConfig) || {}`. 

I prefer that because, similar `try`/`catch`, IMO it's best to have assertions act on as small a range as possible. They're an inherently dangerous, use-at-your-own-risk feature. 

The end result is the same here: `RulesConfig | {}` _(note: that's the union type, not a `||` typo)_ collapses down to just `RulesConfig` in the type system.

package.json
```diff
@@ -8,6 +8,7 @@
     "build": "node scripts/build.js",
     "build:readme": "node tools/update-readme.js",
     "lint": "eslint .",
+    "tsc": "tsc",
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I'm not sure that this fits any of the existing molds. Walking through my thought process:

1. **Lint**: I think type checking is closest to _"scripts that statically analyze files"_. But, TypeScript is not a linter, and users tend to get confused about the difference. I've seen confusion arise directly from labeling it as such.
2. **Test**: TypeScript doesn't execute code, and that's another (less) common confusion among users I've seen.

Suggestion: maybe `"types"` if you're looking for a more tool-agnostic name?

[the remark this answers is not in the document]
nzakas: While I understand your analysis, I don't think this warrants a new script name. If running `tsc` is just meant to verify the types are correct, then `test:types` describes it perfectly.

packages/plugin-kit/src/config-comment-parser.js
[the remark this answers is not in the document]
nzakas: Thanks for the explanation. I guess I was reading this as "assign a value that is a RulesConfig or an empty object", meaning that the type of `items` isn't guaranteed to be `RulesConfig`. Maybe that's okay?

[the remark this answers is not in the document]
nzakas: Ah gotcha. Thanks for explaining 👍

AuntyEms (approved):

nzakas:
@JoshuaKGoldberg are you still working on this?

JoshuaKGoldberg:
Yes! I'd been waiting on #118 to get resolved. Cleaning this up now.

fasttime:
@JoshuaKGoldberg there's a merge conflict now.

.github/workflows/ci.yml
```diff
@@ -9,6 +9,24 @@ on:
...
+            - uses: actions/checkout@v4
+
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: "lts/*"
+
+            - name: Install Packages
+              run: |
+                  npm install
+                  npm run build
```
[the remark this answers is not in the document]
nzakas: The types definition files aren't there until we build.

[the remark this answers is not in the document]
harish-sethuraman: If that is so, doesn't package build step fail when the type is wrong for the package? 🤔 do we need another tsc command to run separately on the main repo?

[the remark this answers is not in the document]
JoshuaKGoldberg: Yeah that is true. That's a pain I've just dealt with in other repos. In theory we could do a strategy like `|| 0` to ignore any failures.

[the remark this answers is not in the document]
harish-sethuraman: just to re iterate, my doubt was why we'd need to run `build` command and `test:types` on source files? 
- the build process itself has a in-built `tsc` check? if the build fails then `tsc` is also expected to fail? 
- we cannot have a `tsc` check separately because its dependent on `build` (which in itself has a `tsc` check?)

now the question resolves to do we have any other `ts` files apart from the `ts` files that are built in the repo? if yes then its better to have a `tsc` for fallback

(sorry for the confusion 😅)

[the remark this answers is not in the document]
nzakas: That's a good point. I hadn't noticed that `test:types` just runs `tsc`...for some reason I had it in my head that there was something else running there.

So yes, `npm run build` actually runs `tsc`, so we don't need something separate running `tsc`

fasttime:
@JoshuaKGoldberg just a reminder that there are some comments for you to address in this PR. There's also a merge conflict.

.github/workflows/ci.yml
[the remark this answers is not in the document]
JoshuaKGoldberg: Just clarifying - is there a request for changes?

JoshuaKGoldberg:
👍 on the merge conflict, fixed. I'm not clear on what else I need to do - asked inline. Thanks for the ping.

.github/workflows/ci.yml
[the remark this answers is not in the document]
nzakas: Yes, there's no need to run `tsc` separately because that's already happening via `npm run build`.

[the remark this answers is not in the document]
JoshuaKGoldberg: Ah got it. Thanks, I see now - sorry for giving you the runaround harish-sethuraman.

nzakas (approved):
LGTM. Would like @fasttime to review before merging.

harish-sethuraman (approved):
LGTM. Thanks for clarifying regarding the build / typecheck step Josh

fasttime (approved):
LGTM, thanks!

PR #144, opened 2024-12-30
feat: allow unsafely customise META_FIELDS

yidingww:
#### Prerequisites checklist

Currently config array assumes `name` is the only meta field that should be treated as descriptive information when checking ignore and matching configs.

However, in certain scenarios, we should allow customization of the default `META_FIELDS`.

Expose an `UNSAFE_extraMetaFields`, which will be combined with `const META_FIELDS` as the instance's `this.metaFields`, which will be used everywhere instead of `const META_FIELDS`.

#### Related Issues

N/A

N/A

### DIFF
--- a/packages/config-array/src/config-array.js
+++ b/packages/config-array/src/config-array.js
@@ -72,10 +72,10 @@ const MINIMATCH_OPTIONS = {
 const CONFIG_TYPES = new Set(["array", "function"]);
 
 /**
- * Fields that are considered metadata and not part of the config object.
+ * Default fields that are considered metadata and not part of the config object.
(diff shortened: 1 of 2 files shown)


nzakas:
I'm sorry, I can't tell what use case you're trying to solve. In any case, this isn't the way we design APIs. If you have a use case in mind, please open an issue and describe it so we can discuss.

comp615:
@nzakas this probably should have included a reference to the upstream patch where the author is trying to update config-inspector so that it more closely mirrors the REAL logic of ESlint: https://github.com/eslint/config-inspector/pull/115

PR #224, opened 2025-06-09
Chore move cli tools to root level

lumirlumir:
#### Prerequisites checklist

#### Related Issues

Reference: https://github.com/eslint/js/pull/650

### DIFF
--- a/packages/config-array/package.json
+++ b/packages/config-array/package.json
@@ -57,8 +57,7 @@
     "c8": "^9.1.0",
     "mocha": "^10.4.0",
     "rollup": "^4.16.2",
-    "rollup-plugin-copy": "^3.5.0",
-    "typescript": "^5.4.5"
+    "rollup-plugin-copy": "^3.5.0"
   },
   "engines": {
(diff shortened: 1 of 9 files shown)


packages/config-array/package.json
```diff
@@ -54,11 +54,7 @@
   "devDependencies": {
     "@jsr/std__path": "^1.0.4",
     "@types/minimatch": "^3.0.5",
-    "c8": "^9.1.0",
-    "mocha": "^10.4.0",
-    "rollup": "^4.16.2",
-    "rollup-plugin-copy": "^3.5.0",
-    "typescript": "^5.4.5"
+    "rollup-plugin-copy": "^3.5.0"
```
nzakas, line 58: Should we also move `rollup-plugin-copy` to the root? It's used in a bunch of the packages.

lumirlumir: Thanks for your suggestion. However, unfortunately, if we remove `rollup-plugin-copy` from the devDependencies, it shows an ESLint error.

image

PR #231, opened 2025-06-23
fix: Add missing peer dependencies to `@eslint/config-helpers`

Janpot:
#### Prerequisites checklist

Fix types resolving issues in my project. It's resolving to an outdated `@types/eslint` instead of `eslint`

Copy peer dependencies from `@eslint/compat`

### DIFF
--- a/packages/config-helpers/package.json
+++ b/packages/config-helpers/package.json
@@ -51,6 +51,14 @@
     "eslint": "^9.27.0",
     "rollup-plugin-copy": "^3.5.0"
   },
+  "peerDependencies": {
+    "eslint": "^8.40 || 9"
+  },
+  "peerDependenciesMeta": {
+    "eslint": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
   }


nzakas:
Thanks for the PR. Please describe the issue you're seeing so we can understand the context of this change.

Janpot:
Sure, `@eslint/config-helpers` is referencing `eslint` types in its [output](https://github.com/eslint/rewrite/blob/40d31ba42a9fe0da10b6ca5e1b10f1f2b10c5f90/packages/config-helpers/src/define-config.js#L505) types, but does not anywhere specify it as a dependency. Under certain conditions, this leads to "The inferred type of "X" cannot be named without a reference to "Y". This is likely not portable. A type annotation is necessary." errors. In our case we're developing a package that exports a shareable config. The package has `eslint` as a peer dependency and also has a transitive dependency on a eslint plugin that depends on `@types/eslint`. Typescript gets confused until we add the correct peer dependency in `@eslint/config`.

nzakas:
Thanks, so this is actually the same issue that was reported here:
https://github.com/eslint/rewrite/issues/226

We can't actually add a peerdependency to `eslint` on this package (see the discussion). We're working on another solution.

sam3k:
We've agreed to not accept the PR on 2025-06-26 TSC meeting and instead move forward with moving types into @eslint/core to solve this problem. @nzakas will prioritize working on this in the next couple of weeks.

PR #248, opened 2025-07-28
fix: Allow MCP server to fix issues when asked to

### LINKED ISSUE #246: Change Request: [MCP] Optionally remove the "Do not automatically fix these issues" trailer

- [ ] `@eslint/compat`
- [ ] `@eslint/config-array`
- [ ] `@eslint/core`
- [x] `@eslint/mcp`
- [ ] `@eslint/object-schema`
- [ ] `@eslint/plugin-kit`

Putting instructions to the agent in the response in the form of "Do not automatically fix these issues..." or any other instructions isn't ideal - it might conflict with instructions that the user has specified.  For example, I provide my claude code with the eslint MCP *precisely* so that it can fix issues that eslint finds as it goes. Because of this instruction on every response from the MCP server, the agent can get very confused.

Instructions for the agent properly belong *outside* of any MCP response. If you want to compose the MCP response with instructions not to automatically fix, (or conversely *TO* automatically fix) then you should do that elsewhere in your agent's instructions/system prompt/AGENTS.md/etc.

Perhaps a good solution would be to have a CLI flag to disable this? I think conceptually it'd just be better to not have it there at all, but disabling for people who DO want to auto-fix (or want other follow-up by the agent) would also work.

### Participation

### Additional comments

I could probably write a pretty quick fix across mcp-server.js and the CLI to pass an arg, but not sure what the "standard" way of doing that in the project would be, etc. If you think it's a good idea and would like me to take a stab at it, I can put the time in.

nzakas:
#### Prerequisites checklist

Make it so the MCP server will allow agents to automatically fix lint errors when the user has provided those instructions.

- Changed the prompt in the `lint-files` tool to state that the lint issues can be fixed if the user has instruction the agent to do so.
- Also updated the `lint-files` result so that it will always output the lint results regardless if fixes were applied. (By default, it seemed like the agent would not output the full results because it would prioritize fixing the found issues over displaying issues it was fixing.)

#### Related Issues

fixes #246

### DIFF
--- a/packages/mcp/src/mcp-server.js
+++ b/packages/mcp/src/mcp-server.js
@@ -48,11 +48,11 @@ mcpServer.tool(
 
 		content.unshift({
 			type,
-			text: "Here are the results of running ESLint on the provided files:",
+			text: "Here are the results of running ESLint on the provided files. Before doing anything else, you must display the full list to the user:",
 		});
 		content.push({
 			type,
-			text: "Do not automatically fix these issues. You must ask the user for confirmation before attempting to fix the issues found.",
+			text: "If the user asked to fix any issues found, proceed in fixing them. If the user did not ask to fix issues found, you must ask the user for confirmation before attempting to fix the issues found.",
 		});
 
 		return {


nzakas:
@hughescr Can you try out this change and see if it works for you?

I think this should work: `npx 'https://gitpkg.vercel.app/eslint/rewrite/packages/mcp?issue246'

hughescr:
Thanks - the concept looks good; let me try it out and lyk

hughescr:
I had to switch to `https://gitpkg.vercel.app/eslint/rewrite/packages/mcp?main` (`?issue246` didn't work), but it's going well so far after a day of use.

hughescr:
Thanks for the quick fix!

PR #297, opened 2025-10-14
feat!: Require Node.js ^20.19.0 || ^22.13.0 || >=24

aladdin-add:
#### Prerequisites checklist

- [ ] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

Updated package.json of the package to require:

This drops support for Node.js 18.

#### Related Issues
refs https://github.com/eslint/eslint/issues/19969

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "!(*.{js,ts})": "prettier --write --ignore-unknown"
   },
   "engines": {
-    "node": "^20.19.0 || ^22.13.0 || >=24"
+    "node": "^22.13.0 || >=24"
   },
   "devDependencies": {
     "@eslint/config-helpers": "file:packages/config-helpers",


mdjermanovic (commented):
Can you also update the template for new projects?

https://github.com/eslint/rewrite/blob/27fdbe7c62d67f5807402e734f4b9b584585e94d/templates/package/package.json#L54-L56

mdjermanovic (approved):
LGTM, thanks!

SwetaTanwar:
LGTM, thanks!

mdjermanovic:
> I believe the `engines` field was first added in [973144e](https://github.com/eslint/rewrite/commit/973144e5a4b4bf691a8e2b0eed0d4698481e1fac), but it seems fine to set it to `^20.19.0 || ^22.13.0 || >=24`, since all internal Node.js modules used by the `rewrite` package are available with Node.js `^20.19.0`.

This was added because of `fs.cpSync()`, which is used in `tools/new-pkg.js`.

lumirlumir:
I missed that the stable version was released as v22.3.0 👍

If so, does using `^22.13.0 || >=24` make sense here? It seems a bit odd for me that the internal package.json and the root-level package.json use different engine versions.

aladdin-add:
> It seems a bit odd for me that the internal package.json and the root-level package.json use different engine versions.

IMHO, it's fine - they are different things - unlike others, the root-level act as something like `devEngines` : https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/blob/main/devengines-field-proposal.md

lumirlumir:
@aladdin-add Thanks for the reference! 

If so, please feel free to revert my suggestion or set it to `^22.13.0 || >=24` depending on the consensus :)

lumirlumir (approved):
LGTM!

PR #298, opened 2025-10-16
fix: require `fix` in suggestion objects

crimsonjay0:
#### Prerequisites checklist

This PR updates TypeScript types for suggestion objects so that the `fix` property is required.

This prevents creating suggestions without a fix, which would otherwise cause the runtime error:
```
TypeError: context.report() called with a suggest option without a fix function.
```

It also broadens the type of `data` from `Record<string, string>` → `Record<string, unknown>` because ESLint message interpolation can accept any value that can be stringified.

- Made `fix` required in `SuggestedEditBase`.
- Changed `data` type to `Record<string, unknown>` for compatibility with ESLint message interpolation.
- Added tests for `data` to ensure various types are allowed.
- Added `@ts-expect-error` in test where `fix: null` is used.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -487,7 +487,7 @@ interface ViolationReportBase {
 	/**
 	 * The data to insert into the message.
 	 */
-	data?: Record<string, string> | undefined;
+	data?: Record<string, unknown> | undefined;
 
 	/**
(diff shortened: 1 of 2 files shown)


DMartens (approved):
Changes LGTM, thanks for spotting this.
Leaving this open for 2 days if someone else wants to take a look.

PR #316, opened 2025-11-11
chore: restore eslint-plugin-jsdoc rules

Pixel998:
#### Prerequisites checklist

This PR restores the eslint-plugin-jsdoc rules that were previously disabled, so that JavaScript files once again receive the intended documentation linting behavior from eslint-config-eslint.

Removed the blanket disable of jsdoc rules and cleaned up the configuration structure in `eslint.config.js`.


packages/config-array/src/config-array.js
```diff
@@ -395,7 +401,7 @@ async function normalize(
 	 * Async iterables cannot be used with the spread operator, so we need to manually
 	 * create the array to return.
 	 */
-	const asyncIterable = await flatTraverse(items);
+	const asyncIterable = flatTraverse(items);
```
JoshuaKGoldberg, line 408: _Technically_, this is a behavior change. It makes me a little nervous to include any in a `chore:`, even if it's a no-op. IMO this is fine as `flatTraverse` isn't `async` and doesn't return a `Promise`. Just calling out as something to keep an eye on.

packages/object-schema/src/merge-strategy.js
```diff
@@ -12,9 +12,9 @@
 export class MergeStrategy {
 	/**
 	 * Merges two keys by overwriting the first with the second.
-	 * @param {*} value1 The value from the first object key.
-	 * @param {*} value2 The value from the second object key.
-	 * @returns {*} The second value.
+	 * @param {any} value1 The value from the first object key.
+	 * @param {any} value2 The value from the second object key.
+	 * @returns {any} The second value.
```
JoshuaKGoldberg: [Bug] `{any}` isn't correct here. This is a generic function so there should be a type parameter:

```suggestion
	 * @template Value
	 * @param {Value} value1 The value from the first object key.
	 * @param {Value} value2 The value from the second object key.
	 * @returns {Value} The second value.
```

Quick reference as I always have to look these up ~ https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template

(here and elsewhere in the PR in functions that should be generic)

packages/object-schema/src/validation-strategy.js
```diff
@@ -12,7 +12,7 @@
 export class ValidationStrategy {
 	/**
 	 * Validates that a value is an array.
-	 * @param {*} value The value to validate.
+	 * @param {any} value The value to validate.
```
JoshuaKGoldberg: [Bug] This won't change anything specifically here, but preferring `unknown` is generally a better practice and leads to safer default behaviors:

```suggestion
	 * @param {unknown} value The value to validate.
```

(here and elsewhere in the file)

packages/config-array/src/config-array.js
```diff
@@ -160,7 +161,7 @@ function rethrowConfigError(config, index, error) {
 /**
  * Shorthand for checking if a value is a string.
  * @param {any} value The value to check.
- * @returns {boolean} True if a string, false if not.
+ * @returns {value is string} True if a string, false if not.
```
JoshuaKGoldberg, line 168: [Praise] Nice "extra credit" to get the type predicate!

packages/config-array/src/files-and-ignores-schema.js
```diff
@@ -16,7 +16,7 @@
 
 /**
  * Asserts that a given value is an array.
- * @param {*} value The value to check.
+ * @param {any} value The value to check.
```
JoshuaKGoldberg: [Bug] This won't change anything specifically here, but preferring `unknown` is generally a better practice and leads to safer default behaviors:

```suggestion
 * @param {unknown} value The value to check.
```

(here and elsewhere that uses `any` for an unknown-typed value)

eslint.config.js
JoshuaKGoldberg, line 1: [Praise] Nice cleanup here on removing the `...`s, etc. too 🙂

JoshuaKGoldberg (changes requested):
This is lovely, thank you for the cleanup!

I have a couple of touchup requests, but generally am very 👍 on the direction of changes. These'll be very nice for understanding the code better.

eslint.config.js
```diff
@@ -1,43 +1,40 @@
...
-
+	globalIgnores(["**/tests/fixtures/"], "rewrite/global-ignores"),
 	{
+		name: "rewrite/js",
+		files: ["**/*.js"],
 		extends: [eslintConfigESLint],
+		settings: {
+			jsdoc: {
+				preferredTypes: {
+					object: "object",
```
DMartens, line 27:
```suggestion
					Object: "object",
```
This seems to be a typo (we want to disallow `Object` and prefer `object`).

packages/plugin-kit/src/source-code.js
```diff
@@ -634,6 +634,7 @@ export class TextSourceCodeBase {
 
 	/**
 	 * Traverse the source code and return the steps that were taken.
+	 * @abstract
```
DMartens, line 638: Good catch. I think `getParent` should also be marked as `@abstract`.

packages/config-array/tests/config-array.test.js
```diff
@@ -298,6 +303,15 @@ describe("ConfigArray", () => {
 	});
 
 	describe("Validation", () => {
+		/**
+		 * Defines tests that assert validation errors in both async and sync normalize flows.
+		 * @param {Object} options Test options.
+		 * @param {boolean} [options.only=false] If true, run these tests exclusively.
+		 * @param {string} options.title Test title prefix.
+		 * @param {Iterable|Function|Object} options.configs Configs to pass to the `ConfigArray` constructor.
```
DMartens:
```suggestion
		 * @param {Iterable<Function|Object>|Function|Object} options.configs Configs to pass to the `ConfigArray` constructor.
```
The `Iterable` type should have type arguments.

eslint.config.js
Pixel998: This isn't a typo. It's overriding the `preferredTypes.object` mapping from `eslint-config-eslint` so `object` is no longer replaced with `Object`. I should probably file an issue in the eslint repo to update `eslint-config-eslint`.

DMartens: Thanks for the clarification. I agree that this should be fixed upstream as in my opinion this should be reversed (`object` should be preferred over `Object`).

DMartens (dismissed):
Changes LGTM, thanks. Leaving open for @JoshuaKGoldberg.

DMartens (approved):

DMartens:
As Joshua said he currently does not have the bandwith for open source, I'm going to merge this.

PR #327, opened 2025-11-17
fix: make `data` property stricter

lumirlumir:
## Prerequisites checklist

## What is the purpose of this pull request?

In this PR, I've resolved the issue metioned in #310.

I've updated the `data` property to accept `string | number | boolean | bigint | null | undefined` and added test cases to ensure it works.

## What changes did you make? (Give an overview)

In this PR, I've resolved the issue metioned in #310.

## Related Issues

Closes: #310 

## Is there anything you'd like reviewers to focus on?

N/A


packages/core/src/types.ts
```diff
@@ -451,7 +451,9 @@ export interface ViolationReportBase {
 	/**
 	 * The data to insert into the message.
 	 */
-	data?: Record<string, unknown> | undefined;
+	data?:
```
DMartens: I think rather than the note, we should use a type alias to make sure they are in sync:
`type MessagePlaceholderData = Record<string, string | number | boolean | bigint | null | undefined>` and then reference them for the root and suggestions.

lumirlumir: Yes — this approach makes sense to me as well.

I also checked `eslint`’s types, and it looks like there are two other places this type could be reused:

- https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L1378
- https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L1390

So maybe define a reusable type like the following, export it, and apply it in `eslint` once a new version of `@eslint/core` is released?

```ts
export type MessagePlaceholderData = Record<string, string | number | boolean | bigint | null | undefined> | undefined;
```

Any further thoughts on this @eslint/eslint-team ?

---

If this approach and the type name look reasonable, I can go ahead and open a PR in `eslint` for this change.

nzakas: Makes sense. :+1:

lumirlumir: I've updated it in https://github.com/eslint/rewrite/pull/327/commits/06d6ba0ba44258bac97fb08af011f08d8659e6ab 👍

packages/core/src/types.ts
```diff
@@ -447,11 +447,15 @@ export type RuleFixer = (
 	fixer: RuleTextEditor,
 ) => RuleTextEdit | Iterable<RuleTextEdit> | null;
 
+export type MessagePlaceholderData =
```
DMartens: I think we should not include `| undefined` here but in every reference which are currently only optional properties but other consumers (as this is exported) do not expect this to be `undefined`.

lumirlumir: Maybe, do you mean using the following type without `undefined`?

```diff
- export type MessagePlaceholderData = Record<string, string | number | boolean | bigint | null | undefined> | undefined;
+ export type MessagePlaceholderData = Record<string, string | number | boolean | bigint | null | undefined>;
```

I don't have a strong opinion on this, but regarding the comment https://github.com/eslint/rewrite/pull/327#discussion_r2549952338, there appears to be consensus in favor of the current (former) type, so let's wait a bit to gather more input.

image

fasttime: I agree with @DMartens it's better not no include `undefined` in the definition of `MessagePlaceholderData`. We use this pattern consistently in other cases, even when a type is always referenced in union with `undefined`, e.g.:

https://github.com/eslint/rewrite/blob/0455323682227ba2e219645a49c20085ab76cbf0/packages/core/src/types.ts#L969

For third-party consumers of a type, it's easier to add `undefined` explicitly where necessary than excluding it from the union when it's unexpected.

lumirlumir: Thanks everyone for the suggestions. I've updated it here: https://github.com/eslint/rewrite/pull/327/commits/e31618053d016c8f06f923869dc530634d4bf083

lumirlumir:
It seems that the CI failure is unrelated to this change.

lumirlumir:
I investigated this and it appears to be a regression in `@modelcontextprotocol/sdk@1.23.0`, released yesterday. Version `1.23.0` added support for Zod v4, which seems to have caused errors in `eslint/mcp`.

I tried addressing it but couldn't find a clear solution, so I opened an issue in the `@modelcontextprotocol/sdk` repo: https://github.com/modelcontextprotocol/typescript-sdk/issues/1180

packages/mcp/package.json
```diff
@@ -36,7 +36,7 @@
     "node": "^20.19.0 || ^22.13.0 || >=24"
   },
   "dependencies": {
-    "@modelcontextprotocol/sdk": "^1.22.0",
+    "@modelcontextprotocol/sdk": "1.22.0",
```
lumirlumir: I've pinned the version as a workaround for the CI failure.

(I tried using `@ts-expect-error` on the exact line where the error occurred, but it didn't work well.)

package.json
```diff
@@ -42,7 +42,7 @@
     "knip": "^5.62.0",
     "lint-staged": "^16.0.0",
     "mocha": "^11.5.0",
-    "prettier": "^3.4.1",
+    "prettier": "^3.7.0",
```
lumirlumir: Meanwhile, `prettier@3.7.0` was released 30 minutes ago and has also caused CI to fail.

Some additional code diffs were introduced by this change.

lumirlumir:
Marking it as a draft until https://github.com/eslint/rewrite/pull/332 is merged.

lumirlumir:
This PR is now ready for re-review.

packages/core/src/types.ts
```diff
@@ -447,11 +447,16 @@ export type RuleFixer = (
 	fixer: RuleTextEditor,
 ) => RuleTextEdit | Iterable<RuleTextEdit> | null;
 
+export type MessagePlaceholderData = Record<
+	string,
+	string | number | boolean | bigint | null | undefined
+>;
```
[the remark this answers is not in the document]
lumirlumir: Make sense to me. Updated.

PR #348, opened 2025-12-31
fix: tighten types of `MergeStrategy` in `object-schema`

lumirlumir:
## Prerequisites checklist

- [ ] `@eslint/compat`
- [ ] `@eslint/config-array`
- [ ] `@eslint/core`
- [ ] `@eslint/mcp`
- [x] `@eslint/object-schema`
- [ ] `@eslint/plugin-kit`

In the public API of `@eslint/object-schema`, the return types of `MergeStrategy.overwrite`, `MergeStrategy.replace`, and `MergeStrategy.assign` are currently typed as `any`, which feels too broad and could be narrowed to a clearer type.

For example:

`MergeStrategy.overwrite` always returns `value2`; `MergeStrategy.replace` always returns either `value1` or `value2`; and `MergeStrategy.assign` always returns the merged object produced by `Object.assign()`.

I think if we use a template for these static functions, we can make the types clearer.

## What changes did you make? (Give an overview)

In this PR, I've tightened the types for `MergeStrategy` in `object-schema`.

The previous return type `any` was too broad, and from a TypeScript perspective, using `any` is generally not recommended.

By using templates, the return types can be narrowed, so I went ahead with this approach.

## Related Issues

N/A

## Is there anything you'd like reviewers to focus on?

N/A

### DIFF
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -38,7 +38,7 @@
     "node": "^20.19.0 || ^22.13.0 || >=24"
   },
   "dependencies": {
-    "@modelcontextprotocol/sdk": "1.25.3",
+    "@modelcontextprotocol/sdk": "^1.26.0",
     "eslint": "^10.0.0",
     "zod": "^4.1.13"
   },
--- a/packages/mcp/tests/mcp-server.test.js
+++ b/packages/mcp/tests/mcp-server.test.js
@@ -60,6 +60,10 @@ describe("MCP Server", () => {
 		await client.connect(clientTransport);
 	});
 
+	afterEach(async () => {
+		await mcpServer.close();
+	});
+
 	describe("Tools", () => {
 		it("should list tools", async () => {
 			const { tools } = await client.listTools();


packages/object-schema/src/merge-strategy.js
```diff
@@ -23,9 +25,11 @@ export class MergeStrategy {
...
 	 * Merges two keys by replacing the first with the second only if the
 	 * second is defined.
-	 * @param {*} value1 The value from the first object key.
-	 * @param {*} value2 The value from the second object key.
-	 * @returns {*} The second value if it is defined.
+	 * @template TValue1 The type of the value from the first object key.
+	 * @template TValue2 The type of the value from the second object key.
+	 * @param {TValue1} value1 The value from the first object key.
+	 * @param {TValue2} value2 The value from the second object key.
+	 * @returns {TValue1 | TValue2} The second value if it is defined.
```
lumirlumir, line 32: Since the `replace` static method always returns `value2` when `value2` is not `undefined`, the return type could be narrowed.

However, there are edge cases where `value2` may be, for example, `string | undefined`, and I don’t see a clear way to express a more specific type for this function right now.

Using `TValue1 | TValue2` as the return type seems like the most general approach here. If there’s a way to narrow it more precisely, I’m happy to follow it.

packages/object-schema/src/merge-strategy.js
```diff
@@ -37,9 +41,11 @@ export class MergeStrategy {
 
 	/**
 	 * Merges two properties by assigning properties from the second to the first.
-	 * @param {*} value1 The value from the first object key.
-	 * @param {*} value2 The value from the second object key.
-	 * @returns {*} A new object containing properties from both value1 and
+	 * @template {Record<string | number | symbol, unknown>} TValue1 The type of the value from the first object key.
+	 * @template {Record<string | number | symbol, unknown>} TValue2 The type of the value from the second object key.
```
lumirlumir: According to the JSDoc comments and the `README.md`, it seems the restriction on `value1` and `value2` is that they must be objects with properties, but I'm not entirely sure.

https://github.com/eslint/rewrite/blob/b65204d085e40425f9ee62c06e9b5005e23651cc/packages/object-schema/src/merge-strategy.js#L39

https://github.com/eslint/rewrite/blob/b65204d085e40425f9ee62c06e9b5005e23651cc/packages/object-schema/README.md?plain=1#L97

packages/object-schema/tests/types/types.test.ts
```diff
@@ -8,3 +8,81 @@
...
+MergeStrategy.overwrite("a", "b") satisfies boolean;
+// @ts-expect-error Type 'boolean' does not satisfy the expected type 'number'.
+MergeStrategy.overwrite(true, false) satisfies number;
+// @ts-expect-error Type 'boolean' does not satisfy the expected type 'string'.
+MergeStrategy.overwrite(true, false) satisfies string;
+
+MergeStrategy.replace(1, 2) satisfies number;
+MergeStrategy.replace("a", "b") satisfies string;
+MergeStrategy.replace(true, false) satisfies boolean;
+MergeStrategy.replace({ a: 1 }, { b: 2 }) satisfies Record<string, number>;
```
lumirlumir, line 80: I intentionally wrote this type test a little loosely, per the https://github.com/eslint/rewrite/pull/348#discussion_r2657805371, in case the types can be narrowed further in the future.

For example, we can use the following test case instead to ensure its behavior:

```diff
- MergeStrategy.replace(1, 2) satisfies number
+ MergeStrategy.replace(1, 2) satisfies 1 | 2
```

(FYI: this type test is also ensured and fails if we revert this change.)

However, If narrowing the type test is needed here as well, I'm happy to do that.

packages/object-schema/src/merge-strategy.js
fasttime: It looks like `value1` can be `undefined`. This is the relevant test case:

https://github.com/eslint/rewrite/blob/72551007b36c63e05e5005536da821be070cdca1/packages/object-schema/tests/object-schema.test.js#L264-L274

Stepping through with the debugger shows that `MergeStrategy.assign()` is invoked twice: 
1. with `value1 = undefined` and `value2 = { bar: true }`  
1. with `value1 = { bar: true }` and `value2 = { baz: false }`

lumirlumir: Sorry for the delay.

I've updated it at https://github.com/eslint/rewrite/pull/348/changes/26d4892c6f8b4059a24dea74cddeabd834824f03, but I found that the current `tsconfig.json` for the type tests doesn't enable `strict` and `exactOptionalPropertyTypes`, so the test I added still passes even after I revert the change.

https://github.com/eslint/rewrite/blob/4d4709b46a4c43432a57554eef255e2df5cceb81/packages/object-schema/tests/types/tsconfig.json#L1-L8

However, if I enable `strict` and `exactOptionalPropertyTypes`, unrelated code starts throwing type errors, so I'm not sure what's the best approach. Do you recommend enabling `strict` and `exactOptionalPropertyTypes` in this PR and fixing the resulting type errors here, or merging this PR now and enabling them in a separate PR?

I can work on either option depending on your advice.

fasttime: Oh, I see. I think it's a good idea to enable `strict` and `exactOptionalPropertyTypes` in type tests even if the sources are compiled without those options. Feel free to do that in this PR or in another one as you prefer. We can keep the test in 26d4892c6f8b4059a24dea74cddeabd834824f03 nonetheless.

lumirlumir: Great, thanks! I’ve worked on it, but I think the scope of the change is somewhat different to be included here, so if it’s alright, I’d like to continue the work in the newly opened PR https://github.com/eslint/rewrite/pull/370.

PR #353, opened 2026-01-07
chore: remove `src/types.ts` from knip entry configuration

Pixel998:
#### Prerequisites checklist

This PR configures knip to properly recognize public API types in `src/types.ts` files so they aren't incorrectly flagged as unused exports. These types are meant for external consumption by package users, even if they're not used internally within our codebase.

Added `src/types.ts` to the knip entry configuration for four packages that export public types:
- `@eslint/config-array`
- `@eslint/config-helpers`
- `@eslint/migrate-config`
- `@eslint/plugin-kit`

The `@eslint/object-schema` package already had this configuration.

#### Related Issues

https://github.com/eslint/rewrite/pull/351

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -9,15 +9,11 @@
       "project": ["src/**/*.{cts,js,ts}", "tests/*.{cts,js,ts}"]
     },
     "packages/config-array": {
-      "entry": [
-        "tests/**/*.{cts,js,ts}",
-        "src/types.ts",
-        "rollup.std__path-config.js"
-      ],
+      "entry": ["tests/**/*.{cts,js,ts}", "rollup.std__path-config.js"],
       "project": ["src/**/*.{cts,js,ts}", "tests/**/*.{cts,js,ts}"]
     },
     "packages/config-helpers": {
-      "entry": ["tests/**/*.{cts,js,ts}", "src/types.ts"],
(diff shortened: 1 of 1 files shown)


fasttime:
I believe knip is already recognizing `src/types.ts` files; otherwise it would flag them as unused. The issue in #351 seems to be an unreferenced type definition. I’ve added a comment on the PR with a possible fix.

Pixel998:
Sorry, I meant this knip issue: https://github.com/eslint/rewrite/actions/runs/20762026304/job/59619177443

fasttime:
> Sorry, I meant this knip issue: https://github.com/eslint/rewrite/actions/runs/20762026304/job/59619177443

That's the same error. It will no longer appear now that `CustomRuleVisitorWithExit` is being reexported.

Pixel998:
You're right. My point was that knip needs explicit re-exports to avoid flagging types. I originally opened this PR thinking we'd need a structural fix (like `/types` exports or the `eslint/markdown` dummy file approach) that would require telling knip about `src/types.ts`.

Since we've gone with the re-export approach instead, these configuration entries are no longer needed. I've updated the PR to remove all of them.

fasttime (approved):
LGTM, thanks!

PR #357, opened 2026-01-18
build: standardize test scripts

Pixel998:
#### Prerequisites checklist

Standardizes the test and build orchestration across all packages in the monorepo.

- Introduce `test:unit` script across all packages for granular testing
- Update `test` scripts to run both types and unit tests where applicable
- Update CI workflow to use `test:unit` in the build matrix
- Move package build logic to `pretest` scripts for consistency
- Update package template to reflect new script standards

### DIFF
--- a/packages/compat/package.json
+++ b/packages/compat/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@eslint/compat",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "description": "Compatibility utilities for ESLint",
   "type": "module",
   "main": "dist/esm/index.js",
@@ -52,7 +52,7 @@
   },
   "homepage": "https://github.com/eslint/rewrite/tree/main/packages/compat#readme",
   "dependencies": {
-    "@eslint/core": "^1.0.1"
+    "@eslint/core": "^1.1.0"
   },
   "devDependencies": {
     "@types/node": "^24.7.2",
--- a/packages/config-helpers/package.json
+++ b/packages/config-helpers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@eslint/config-helpers",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "description": "Helper utilities for creating ESLint configuration",
   "type": "module",
   "main": "dist/esm/index.js",
@@ -51,7 +51,7 @@
   },
   "homepage": "https://github.com/eslint/rewrite/tree/main/packages/config-helpers#readme",
   "dependencies": {
-    "@eslint/core": "^1.0.1"
+    "@eslint/core": "^1.1.0"
   },
   "devDependencies": {
     "rollup-plugin-copy": "^3.5.0"
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,10 +1,10 @@
 {
-  "packages/compat": "2.0.1",
+  "packages/compat": "2.0.2",
   "packages/config-array": "0.23.0",
-  "packages/config-helpers": "0.5.1",
-  "packages/core": "1.0.1",
+  "packages/config-helpers": "0.5.2",
+  "packages/core": "1.1.0",
   "packages/mcp": "0.2.0",
-  "packages/migrate-config": "2.0.1",
+  "packages/migrate-config": "2.0.2",
   "packages/object-schema": "3.0.0",
-  "packages/plugin-kit": "0.5.1"
+  "packages/plugin-kit": "0.6.0"
 }
--- a/packages/compat/CHANGELOG.md
+++ b/packages/compat/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [2.0.2](https://github.com/eslint/rewrite/compare/compat-v2.0.1...compat-v2.0.2) (2026-01-29)
+
+
+### Bug Fixes
+
+* add eslint 10 as peer dependency ([#361](https://github.com/eslint/rewrite/issues/361)) ([ecb37dc](https://github.com/eslint/rewrite/commit/ecb37dcafc6513649c03f245f0f2505e7eb10dd1))
+
+
+### Dependencies
+
+* The following workspace dependencies were updated
+  * dependencies
+    * @eslint/core bumped from ^1.0.1 to ^1.1.0
+
 ## [2.0.1](https://github.com/eslint/rewrite/compare/compat-v2.0.0...compat-v2.0.1) (2026-01-08)
 
 
--- a/packages/compat/jsr.json
+++ b/packages/compat/jsr.json
@@ -1,6 +1,6 @@
 {
   "name": "@eslint/compat",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "exports": "./dist/esm/index.js",
   "publish": {
     "include": [
--- a/packages/config-helpers/CHANGELOG.md
+++ b/packages/config-helpers/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [0.5.2](https://github.com/eslint/rewrite/compare/config-helpers-v0.5.1...config-helpers-v0.5.2) (2026-01-29)
+
+
+### Dependencies
+
+* The following workspace dependencies were updated
+  * dependencies
+    * @eslint/core bumped from ^1.0.1 to ^1.1.0
+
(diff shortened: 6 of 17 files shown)


fasttime:
The `test:types` scripts are also run in CI at time when the whole project has been built, so I'm not sure If we'd want to provide separate scripts to run the type tests with and without a build first.

https://github.com/eslint/rewrite/blob/1280e7987aef36e3a90253574232be585b8e1eef/.github/workflows/ci.yml#L74-L97

lumirlumir (commented):
I think this concerns the common convention across repositories that include a `test:types` script.

In the CSS repository, the `tsc -p tests/types/tsconfig.json` script appears to run after `npm run build` 

(ref: https://github.com/eslint/css/blob/321d4705e396f0838e7960de45c8dd5947e2ecf1/package.json#L62).

However, aside from the CSS repository, I don't see other repositories running `npm run build` before testing types.

I don't have a strong preference, but it would be nice to standardize this across repositories.

fasttime:
In `eslint/js`, the type tests of each package are included when running the `test` script, and the `test` script builds the package through `pretest` if necessary. E.g. in `espree`:

```json
    "pretest": "npm run build",
    "test": "npm run test:types && npm run test:cjs && npm run test:esm",
    "test:cjs": "mocha --color --reporter progress --timeout 30000 \"tests/**/*.test.cjs\"",
    "test:esm": "c8 mocha --color --reporter progress --timeout 30000 \"tests/**/*.test.js\"",
    "test:types": "tsd --typings dist/espree.d.ts"

```

https://github.com/eslint/js/blob/98caac0d92865a014220cf9e014df08f8913a4a4/packages/espree/package.json#L66-L70

If the intent is simplifying entering commands during development, perhaps this could be an option?

lumirlumir:
> If the intent is simplifying entering commands during development, perhaps this could be an option?

I think it's a good option, since CSS and JSON also use the `pretest` lifecycle hook.

- https://github.com/eslint/json/blob/main/package.json#L49
- https://github.com/eslint/css/blob/main/package.json#L53

One more thing to discuss: unlike `eslint/js`, some repositories have commands under the `test:` scope such as `test:coverage` and `test:jsr`. Would it make sense to group those as well? Their nature differs somewhat from unit and type testing, so I'm not sure what the best approach is.

Maybe it would make sense to group only the type and unit tests into a single `test` script (excluding `test:coverage` and `test:jsr`) and add a `pretest` lifecycle hook?

To clarify, my suggestion is:

```json
    "pretest": "npm run build",
    "test": "npm run test:types && npm run test:unit",
    "test:coverage": "c8 npm test",
    "test:jsr": "npx -y jsr@latest publish --dry-run",
    "test:types": "tsc -p tests/types/tsconfig.json"
    "test:unit": "mocha \"tests/**/*.test.js\" --timeout 30000",
```

sethamus:
I've hit this in our other repos too and it’s definitely a friction point—it's easy to waste time debugging a 'fail' that’s actually just a stale build.

I get the concern about CI redundancy, but maybe we could add the `pretest` hook for consistency and then a `test:types:build` script for local dev? That way we keep the CI lean but still have a one-step way to run standalone type tests without the whole suite.

Pixel998:
> Maybe it would make sense to group only the type and unit tests into a single test script (excluding `test:coverage` and `test:jsr`) and add a `pretest` lifecycle hook?

Sounds good to me. @fasttime, what do you think about moving in this direction?

fasttime:
> > Maybe it would make sense to group only the type and unit tests into a single test script (excluding `test:coverage` and `test:jsr`) and add a `pretest` lifecycle hook?
> 
> Sounds good to me. @fasttime, what do you think about moving in this direction?

I think it's still good to run unit tests separately from type tests in GitHub Actions. Currently, this is done by running the top-level `test` script from the workflow file, which in turn runs `npm test` on each package:

https://github.com/eslint/rewrite/blob/11e1df2d30733f1b8db67a5ffdb05ee1f005e505/.github/workflows/ci.yml#L51-L53

https://github.com/eslint/rewrite/blob/11e1df2d30733f1b8db67a5ffdb05ee1f005e505/package.json#L7

If the `test` scripts in the packages are changed to include type tests, how will we run unit tests in CI? Shall we add a top-level `test:unit` script for that?

Pixel998:
> If the test scripts in the packages are changed to include type tests, how will we run unit tests in CI? Shall we add a top-level test:unit script for that?

Sounds good 👍. We can add a top-level `test:unit` script:
```json
"test:unit": "npm run test:unit --workspaces --if-present"
```

fasttime:
> > If the test scripts in the packages are changed to include type tests, how will we run unit tests in CI? Shall we add a top-level test:unit script for that?
> 
> Sounds good 👍. We can add a top-level `test:unit` script:
> 
> ```json
> "test:unit": "npm run test:unit --workspaces --if-present"
> ```

That should work. Noting that `npm test` is also run in `test:coverage` scripts. To keep the current behavior there, those should be updated as well to run (I think) `npm run build` and then `npm run test:unit`.

packages/compat/package.json
```diff
@@ -28,9 +28,11 @@
     "build:cts": "node ../../tools/build-cts.js dist/esm/index.d.ts dist/cjs/index.d.cts",
     "build": "rollup -c && tsc -p tsconfig.esm.json && npm run build:cts",
     "lint:types": "attw --pack",
-    "test": "mocha \"tests/**/*.test.js\"",
-    "test:coverage": "c8 npm test",
-    "test:jsr": "npx -y jsr@latest publish --dry-run"
+    "pretest": "npm run build",
+    "test": "npm run test:unit",
+    "test:coverage": "npm run build && c8 npm run test:unit",
```
fasttime:
```suggestion
    "test:coverage": "c8 npm run test:unit",
```

To maintain the current behavior, I think we don't need to add a build step before coverage. It looks like the `build` script in this package is only useful to validate the types; it's not needed for the unit tests.

packages/mcp/package.json
```diff
@@ -15,8 +15,10 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "mocha \"tests/**/*.test.js\"",
-    "test:coverage": "c8 npm test"
+    "pretest": "npm run build",
+    "test": "npm run test:unit",
+    "test:coverage": "npm run build && c8 npm run test:unit",
```
fasttime: To maintain the current behavior, there should be no build before coverage tests here.

```suggestion
    "test:coverage": "c8 npm run test:unit",
```

packages/object-schema/package.json
```diff
@@ -29,10 +29,12 @@
     "build:cts": "node ../../tools/build-cts.js dist/esm/index.d.ts dist/cjs/index.d.cts",
     "build": "rollup -c && npm run build:dedupe-types && tsc -p tsconfig.esm.json && npm run build:cts",
     "lint:types": "attw --pack",
-    "test": "mocha \"tests/**/*.test.js\"",
-    "test:coverage": "c8 npm test",
+    "pretest": "npm run build",
+    "test": "npm run test:types && npm run test:unit",
+    "test:coverage": "npm run build && c8 npm run test:unit",
```
fasttime: Same here:

```suggestion
    "test:coverage": "c8 npm run test:unit",
```

packages/config-helpers/package.json
```diff
@@ -29,11 +29,13 @@
     "build:cts": "node ../../tools/build-cts.js dist/esm/index.d.ts dist/cjs/index.d.cts",
     "build": "rollup -c && npm run build:dedupe-types && tsc -p tsconfig.esm.json && npm run build:cts",
     "lint:types": "attw --pack",
-    "test": "mocha \"tests/**/*.test.js\"",
-    "test:coverage": "c8 npm test",
+    "pretest": "npm run build",
+    "test": "npm run test:types && npm run test:unit",
+    "test:coverage": "npm run build && c8 npm run test:unit",
```
fasttime:
```suggestion
    "test:coverage": "c8 npm run test:unit",
```

This is to maintain the current behavior, since there was no `pretest` script with `npm run build` previously. Also, the unit tests in this package don't use and built files.

packages/compat/package.json
Pixel998: Removed in https://github.com/eslint/rewrite/pull/357/commits/13f10c57b8bd6ef6d7cc9ff58f365a5a94c2d6f5

fasttime (approved):
LGTM, thanks!

PR #372, opened 2026-02-16
fix: align `object-schema` `PropertyDefinition` types with runtime

Pixel998:
#### Prerequisites checklist

Align `@eslint/object-schema` TypeScript types with the documented and runtime behavior (optional `required`, and allowing `schema`-only definitions).

- Make `PropertyDefinition.required` optional.
- Update `PropertyDefinition` typing to allow `schema`-only definitions, while still requiring `merge`/`validate` when `schema` isn’t present.
- Update type tests to cover the relaxed and required cases.

### DIFF
--- a/packages/object-schema/tests/types/types.test.ts
+++ b/packages/object-schema/tests/types/types.test.ts
@@ -192,7 +192,10 @@ const propertyWithSchema: PropertyDefinition = {
 	},
 };
 
-propertyWithSchema.schema satisfies ObjectDefinition | undefined; // `schema` is optional.
+propertyWithSchema.schema satisfies ObjectDefinition;
+
+(property: PropertyDefinition) =>
+	property.schema satisfies ObjectDefinition | undefined;
 
 // @ts-expect-error -- merge and validate are required when schema isn't present
 const propertyMissingMerge: PropertyDefinition = {


packages/object-schema/src/types.ts
```diff
@@ -20,37 +20,70 @@ export type BuiltInValidationStrategy =
...
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- https://github.com/eslint/rewrite/pull/90#discussion_r1687206213
+type CustomMergeStrategy = (target: any, source: any) => any;
+
+/**
+ * Custom validation strategy.
  */
-export interface PropertyDefinition {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- https://github.com/eslint/rewrite/pull/90#discussion_r1687206213
+type CustomValidationStrategy = (value: any) => void;
```
fasttime: It would be good to export these two types along with their corresponding `BuiltIn`s, so that they can be used by implementations.

https://github.com/eslint/rewrite/blob/1d663c7a3ba58d4b22ca8cbc3b078279a98c2cc4/packages/object-schema/src/object-schema.js#L16-L17

packages/object-schema/src/types.ts
```diff
@@ -20,37 +20,70 @@ export type BuiltInValidationStrategy =
...
 	 */
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- https://github.com/eslint/rewrite/pull/90#discussion_r1687206213
-	validate: BuiltInValidationStrategy | ((value: any) => void);
+	validate: BuiltInValidationStrategy | CustomValidationStrategy;
 
 	/**
 	 * The schema for the object value of this property.
 	 */
-	schema?: ObjectDefinition;
+	schema?: never;
```
fasttime: Can we move this property before `merge` for consistency with `PropertyDefinitionWithSchema`?

packages/object-schema/src/types.ts
```diff
@@ -20,37 +20,70 @@ export type BuiltInValidationStrategy =
...
+	/**
+	 * The strategy to merge the property.
+	 */
+	merge?: BuiltInMergeStrategy | CustomMergeStrategy;
+
+	/**
+	 * The strategy to validate the property.
+	 */
+	validate?: BuiltInValidationStrategy | CustomValidationStrategy;
+}
```
fasttime: It wouldn't hurt to export `PropertyDefinitionWithStrategies` and `PropertyDefinitionWithSchema` as well.

https://github.com/eslint/rewrite/blob/1d663c7a3ba58d4b22ca8cbc3b078279a98c2cc4/packages/object-schema/src/object-schema.js#L19

packages/object-schema/tests/types/types.test.ts
```diff
@@ -137,23 +137,28 @@ MergeStrategy.assign({ a: 1 }, { a: "a" }) satisfies {
...
+} satisfies PropertyDefinition;
 
-propertyWithBuiltInStrategies.required satisfies boolean;
 propertyWithBuiltInStrategies.merge satisfies
 	| BuiltInMergeStrategy
 	| ((target: any, source: any) => any);
 propertyWithBuiltInStrategies.validate satisfies
 	| BuiltInValidationStrategy
 	| ((value: any) => void);
+propertyWithRequired.required satisfies boolean;
```
fasttime: These tests are now validating the types of the properties defined in the test objects, no longer those of `PropertyDefinition`, which defeats the original purpose. Would it make sense to use a type annotation on the objects instead of `satisfies` like it was done previously? For example:

```ts
// PropertyDefinition with built-in strategies
const propertyWithBuiltInStrategies: PropertyDefinition = {
	merge: "replace",
	validate: "string",
};

// PropertyDefinition with a required flag
const propertyWithRequired: PropertyDefinition = {
	required: true,
	merge: "replace",
	validate: "string",
};
```

Or another approach to make sure that the types being validating come from `PropertyDefinition`.

Pixel998: Fixed in https://github.com/eslint/rewrite/pull/372/commits/eb16ed8493c72e60b98617de31c7abfab7a44e0c

packages/object-schema/tests/types/types.test.ts
```diff
@@ -190,8 +194,80 @@ const propertyWithSchema: PropertyDefinition = {
 
 propertyWithSchema.schema satisfies ObjectDefinition | undefined; // `schema` is optional.
```
fasttime: It looks like TypeScript understands now that `schema` is not optional here, so this could be simply:

```ts
propertyWithSchema.schema satisfies ObjectDefinition;
```

Maybe we could add a new test for the general case, e.g.

```ts
(property: PropertyDefinition) => property.schema satisfies ObjectDefinition | undefined;
```

Pixel998: Added 👍

fasttime (approved):
LGTM, thanks!

PR #384, opened 2026-02-24
ci: run type tests on TypeScript `5.x` and `5.3.2`

sethamus:
#### Prerequisites checklist

#### AI acknowledgment

This PR adds CI coverage for the minimum supported TypeScript (5.3.2) while still testing against the latest 5.x.

Updated the `test_types` CI job to run on a `typescript: ["5.x", "5.3.2"]` matrix

#### Related Issues

https://github.com/eslint/eslint/issues/19920

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@eslint/config-helpers": "file:packages/config-helpers",
     "@types/mocha": "^10.0.7",
     "@types/node": "^24.9.1",
-    "c8": "^10.1.3",
+    "c8": "^11.0.0",
     "eslint": "^9.35.0",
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
fasttime, line 1: I think we don't need to re-run the build and type tests on TypeScript 5.3 here. Probably it's sufficient to make sure that something like `tsc dist/**/*.{cts,ts}` passes without compiler errors.

sethamus: I’ve pushed an update. Let me know if this is what you had in mind.

PR #385, opened 2026-02-24
fix: tighten extends typing in config-helpers

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR tightens `@eslint/config-helpers` TypeScript typings so invalid `extends` usage (nested `extends`, and `basePath` inside extension objects) is caught at compile time instead of failing later.

- Add/export `ExtensionConfigObject` and use it in `extends` types.
- Make `defineConfig()` require at least one argument.
- Add type tests.

### DIFF
--- a/packages/config-helpers/src/types.ts
+++ b/packages/config-helpers/src/types.ts
@@ -36,7 +36,4 @@ export interface ConfigWithExtends extends ConfigObject {
 	extends?: ExtendsElement[];
 }
 
-export type ConfigWithExtendsArray = [
-	InfiniteArray<ConfigWithExtends>,
-	...InfiniteArray<ConfigWithExtends>[],
-];
+export type ConfigWithExtendsArray = InfiniteArray<ConfigWithExtends>[];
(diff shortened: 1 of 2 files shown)


packages/config-helpers/src/types.ts
```diff
@@ -26,4 +36,7 @@ export interface ConfigWithExtends extends ConfigObject {
 	extends?: ExtendsElement[];
 }
 
-export type ConfigWithExtendsArray = InfiniteArray<ConfigWithExtends>[];
+export type ConfigWithExtendsArray = [
+	InfiniteArray<ConfigWithExtends>,
+	...InfiniteArray<ConfigWithExtends>[],
+];
```
fasttime: It looks like this use case would stop working after this change:

```ts
import { defineConfig } from '@eslint/config-helpers';
import markdown from '@eslint/markdown';

export default defineConfig(...markdown.configs.recommended, {
  linterOptions: { noInlineConfig: true },
});
```

Pixel998: Good catch. I originally made that change to enforce the 'at least one' requirement for `defineConfig()`, but it doesn't play well with spreads. I've reverted it and added a regression test to ensure this use case stays supported.

PR #388, opened 2026-02-25
fix: correct extraConfigTypes validation error messages

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR fixes confusing `extraConfigTypes` validation errors so they reference the correct option name and only list supported values.

- Refer to `extraConfigTypes` (not `configTypes`) in the “at most two items” error.
- Remove the incorrect `"object"` from the invalid-value error and only list `"array"` / `"function"` as supported values.

### DIFF
--- a/packages/config-array/src/config-array.js
+++ b/packages/config-array/src/config-array.js
@@ -633,14 +633,14 @@ function assertNormalized(configArray) {
 function assertExtraConfigTypes(extraConfigTypes) {
 	if (extraConfigTypes.length > 2) {
 		throw new TypeError(
-			"configTypes must be an array with at most two items.",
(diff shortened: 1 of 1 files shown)


DMartens:
Changes LGTM, thanks for spotting this.
I think we should also make the type for `extraConfigTypes` more specific (from "Array<string>" to "Array<'function' | 'array'>").

Pixel998:
> I think we should also make the type for extraConfigTypes more specific (from "Array" to "Array<'function' | 'array'>").

 I think this is already in place. `ExtraConfigType` is defined as `"array" | "function"`.

DMartens:
I am sorry, I checked whether the types match locally which code was out of date.

PR #393, opened 2026-03-02
build: Allow releases from other branches

### LINKED ISSUE #378: Change Request: Allow releasing previous major versions of @eslint/config-array

- [ ] `@eslint/compat`
- [x] `@eslint/config-array`
- [ ] `@eslint/core`
- [ ] `@eslint/mcp`
- [ ] `@eslint/object-schema`
- [ ] `@eslint/plugin-kit`

Right now, we have no way of releasing to previous major release lines of each package in case of security issues.

Manual publishing causes problems because we typically publish with provenance and manual publishes don't have that.

Set up the repo to allow backports for at least `@eslint/config-array` (which we know needs it).

### Participation

### AI acknowledgment

- [x] I did not use AI to generate this issue report.

### Additional comments

This will be a bit tricky because this is a monorepo, so a generic solution likely won't work and we'll have to do something specific for each package we want backports for.

nzakas:
#### Prerequisites checklist

#### AI acknowledgment

Allows releases from branches in this format:

```
release/<package-name>/v<version>
```
So to release off of `@eslint/config-array@0.20`, we could create a branch called `release/config-array/v0.20` and then send pull requests to it.

Updated the `release-please.yml` file to also target release branches.

#### Related Issues

fixes #378

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -55,6 +55,7 @@ jobs:
               if: ${{ steps.release.outputs.releases_created == 'true' }}
               env:
                   STEPS_RELEASE_OUTPUTS: ${{ toJson(steps.release.outputs) }}
+                  GITHUB_REF_NAME: ${{ github.ref_name }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   TWITTER_API_CONSUMER_KEY: ${{ secrets.TWITTER_CONSUMER_KEY }}
                   TWITTER_API_CONSUMER_SECRET: ${{ secrets.TWITTER_CONSUMER_SECRET }}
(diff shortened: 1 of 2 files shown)


fasttime:
> So to release off of `@eslint/config-array@0.20`, we could create a branch called `release/config-array/v0.20` and then send pull requests to it.

That should work as long the new package version numbers don't overlap with existing npm versions. I think we should also make sure that releases from a `release/*` branch are published with a specific tag instead of `latest` (not sure if Release Please handles that automatically).

nzakas:
Updated to include `--tag maintenance` for any non-`main` branches.

fasttime:
Do we want to keep release branches in the repo once a maintenance release is finished, or is it better to delete them and just keep the tags to avoid confusion?

nzakas:
I'd say we keep the branch around until v9.x is EOL, just in case we need another patch.

fasttime:
In today's TSC meeting we agreed to keep the release branches around until v9.x is EOL, in case we need to do more maintenance releases.

PR #416, opened 2026-03-25
fix: re-export ESM types in CommonJS

fasttime:
#### Prerequisites checklist

#### AI acknowledgment

- [X] I did _not_ use AI to generate this PR.

This PR simplifies CommonJS types by re-exporting ESM types.

Most packages in this repo export types both in CommonJS and ESM format, which is necessary for compatibility with both module systems. Today, the CommonJS declarations duplicate the ESM declarations with only small replacements to match import paths. For example, these are the ESM type declarations for `@eslint/plugin-kit@0.6.1`:

* [dist/esm/index.d.ts](https://app.unpkg.com/@eslint/plugin-kit@0.6.1/files/dist/esm/index.d.ts)
* [dist/esm/types.ts](https://app.unpkg.com/@eslint/plugin-kit@0.6.1/files/dist/esm/types.ts)

And these are the corresponding, very similar CommonJS declarations:

* [dist/cjs/index.d.cts](https://app.unpkg.com/@eslint/plugin-kit@0.6.1/files/dist/cjs/index.d.cts)
* [dist/cjs/types.cts](https://app.unpkg.com/@eslint/plugin-kit@0.6.1/files/dist/cjs/types.cts)

To avoid the duplication and reduce the published package size, this PR changes the CommonJS type declarations to one single file that just re-exports the ESM types:

```ts
// dist/cjs/index.d.cts
import type * as types from "../esm/index.js" with { "resolution-mode": "import" };
export = types;
```

This is now possible thanks to the [`resolution-mode`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html#stable-support-resolution-mode-in-import-types) import attribute, which has been supported since TypeScript 5.3, the earliest TypeScript version supported by ESLint.

This change affects all packages that use the `build:cts` script, i.e.:

- `compat`
- `config-array`
- `config-helpers`
- `core`
- `object-schema`
- `plugin-kit`

* Updated `tools/build-cts.js` script re-exports ESM type declarations instead of duplicating them.
* Updated Rollup configurations so that `dist/cjs/types.cts` files are no longer created.
* Updated `build:cts` script in `core` package to use the new script.

#### Related Issues

refs eslint/eslint#19920

### DIFF
--- a/tools/build-cts.js
+++ b/tools/build-cts.js
@@ -1,9 +1,8 @@
 /**
- * @fileoverview Rewrites import expressions for CommonJS compatibility.
- * This script creates "dist/cjs/index.d.cts" by importing types from "dist/esm/index.d.ts" and
- * re-exporting them using CommonJS syntax.
+ * @fileoverview Creates a CommonJS type declaration file that re-exports ESM types.
  *
- *    node tools/build-cts.js /path/to/esm/index.d.ts path/to/cjs/index.d.cts
+ * Usage:
+ *    node tools/build-cts.js /path/to/esm/index.d.ts /path/to/cjs/index.d.cts
  *
  * @author Francesco Trotta
  */


mdjermanovic:
Will this also solve problems we had in https://github.com/eslint/eslint/pull/19843, since both ESM and CJS entry points of `@eslint/core` will return types declared in the same file?

fasttime:
> Will this also solve problems we had in [eslint/eslint#19843](https://github.com/eslint/eslint/pull/19843), since both ESM and CJS entry points of `@eslint/core` will return types declared in the same file?

Now that you mention it, it seems so! I can see that with this change, interface augmentation for rules and settings works also in ESM. For example, currently this config is not producing any errors:

* [**Repro (unpatched)**](https://stackblitz.com/edit/eslint-issue-19843-repro-unpatched?file=eslint.config.ts)

But if the CJS types are patched like in this PR, TypeScript recognizes the type mismatches and reports the expected errors:

* [**Repro (patched)**](https://stackblitz.com/edit/eslint-issue-19843-repro-patched?file=eslint.config.ts)

That means we could resume working on eslint/eslint#19843 once this change is released.

mdjermanovic:
Great! Then, I think we can consider this a fix since it unifies exported types so I'm changing the tag to `fix`. Also to trigger releases.

PR #420, opened 2026-03-29
fix: type-check defaultLanguageOptions against LangOptions

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR ensures `Language.defaultLanguageOptions` is type-checked against a language’s `LangOptions` type instead of the broad `LanguageOptions` base type.

- Updated the `Language` type in `@eslint/core` so `defaultLanguageOptions` uses `Options["LangOptions"]`.
- Added a type test.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1175,7 +1175,7 @@ export interface Language<
 	/**
 	 * Default language options. User-defined options are merged with this object.
 	 */
-	defaultLanguageOptions?: LanguageOptions;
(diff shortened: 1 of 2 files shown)


DMartens:
Changes LGTM, thanks.
I wondered why all the other interface properties used the type parameter.
It seems the reason is that the PR #110, which introduces the type parameter `LanguageOptions`, and the PR #123 adding `defaultLanguageOptions` overlapped and it was simply an oversight.

PR #421, opened 2026-04-01
feat: add `languages` and `docs.dialects` to rule `meta` types

fasttime:
#### Prerequisites checklist

#### AI acknowledgment

- [X] I did _not_ use AI to generate this PR.

Update the core types.

This PR adds the `languages` property to the `RulesMeta` interface (the type used for rule `meta` objects), and `dialects` to `RulesMetaDocs` (the `meta.docs` type), following the design described in the [rule languages RFC](https://github.com/eslint/rfcs/blob/main/designs/2025-rule-languages/README.md). The related change in ESLint has already been merged in eslint/eslint#20571. It also deprecates `language` and `dialects` in `RulesMeta`.

#### Related Issues

refs eslint/eslint#19462

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -130,6 +130,13 @@ export interface RulesMetaDocs {
 	 * Indicates if the rule is frozen (no longer accepting feature requests).
 	 */
 	frozen?: boolean | undefined;
+
+	/**
+	 * The dialects of the languages that the rule is intended to lint.
+	 * @example
+	 * ["JavaScript", "TypeScript"]
+	 */
+	dialects?: string[] | undefined;
 }
 
 /**
@@ -188,11 +195,13 @@ export interface RulesMeta<
 
 	/**
 	 * The language the rule is intended to lint.
+	 * @deprecated Use `languages` instead.
 	 */
 	language?: string;
 
 	/**
 	 * The dialects of `language` that the rule is intended to lint.
+	 * @deprecated Use `docs.dialects` instead.
 	 */
 	dialects?: string[];
 
--- a/packages/core/tests/types/types.test.ts
+++ b/packages/core/tests/types/types.test.ts
@@ -291,6 +291,7 @@ const testRule: RuleDefinition<{
 		fixable: "code",
 		docs: {
 			recommended: true,
+			dialects: ["JavaScript", "TypeScript"],
 		},
 		deprecated: {
 			message: "use something else",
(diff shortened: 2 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -195,6 +195,12 @@ export interface RulesMeta<
 	 * The dialects of `language` that the rule is intended to lint.
 	 */
 	dialects?: string[];
+
+	/**
+	 * Languages supported by this rule in the format `"plugin/language"`.
+	 * Use `"*"` for any language or `"plugin/*"` for any language from a specific plugin.
+	 */
+	languages?: string[] | undefined;
```
mdjermanovic, line 214: Should we maybe also mark the `language` property on line 192 as deprecated, since it was previously intended for the same purpose?

fasttime: Makes sense. Shall we deprecate also `dialects` as indicated in the RFC?

fasttime: Actually if we deprecate `meta.dialects`, we should also add `meta.docs.dialects`, but we could do that as well.

packages/core/src/types.ts
```diff
@@ -195,6 +195,12 @@ export interface RulesMeta<
 	 * The dialects of `language` that the rule is intended to lint.
 	 */
 	dialects?: string[];
+
+	/**
+	 * Languages supported by this rule in the format `"plugin/language"`.
+	 * Use `"*"` for any language or `"plugin/*"` for any language from a specific plugin.
```
lumirlumir, line 210:
```suggestion
	 * Use `"*"` for any language or `"plugin/*"` for any language from a specific plugin.
	 * @example
	 * ["js/js", "markdown/gfm", "json/jsonc", "css/css"]
```

Maybe adding a simple example here would make sense? I understand other JSDoc entries don't have an example section, but since it's a newly added feature, including a short JSDoc example could be helpful for plugin authors.

(The example could vary if other suggestions are available.)

packages/core/src/types.ts
mdjermanovic: > Shall we deprecate also `dialects` as indicated in the RFC?

I think we should.

> Actually if we deprecate `meta.dialects`, we should also add `meta.docs.dialects`, but we could do that as well.

Make sense to me 👍

packages/core/src/types.ts
fasttime: I think that's a good idea. Just removed `"js/ts"` since the default plugin only supports the [`js` language](https://github.com/eslint/eslint/blob/v10.1.0/lib/config/default-config.js#L36-L38) at this time. This could be changed later.

packages/core/src/types.ts
fasttime: All updated in 80386fedbbdbc76ff364f6148e5c2a2d4efe495f.

lumirlumir (approved):
LGTM, thanks!

Since there are already two approvals, I'll go ahead and merge this PR.

PR #430, opened 2026-04-08
feat: Add new `includeIgnoreFile()` to config-helpers

### LINKED ISSUE #329: Change Request: includeIgnoreFile should respect the path to the directory containing the .gitignore

- [x] `@eslint/compat`
- [ ] `@eslint/config-array`
- [ ] `@eslint/core`
- [ ] `@eslint/mcp`
- [ ] `@eslint/object-schema`
- [ ] `@eslint/plugin-kit`

The current includeIgnoreFile only supports the .gitignore in the root directory. .gitignore in subfolders are common monorepo and even single project repo.

Quote from https://git-scm.com/docs/gitignore

> These patterns match relative to the location of the .gitignore file

The following .gitignore in folderA should
```
# ignore `folderA/**/abc`, not `**/abc`
abc

# ignore `folderA/build`, not `build`
/build
```

The following code is a rough fix. `ignoreFilePath` is relative to the folder containing the config file (not sure if this is the convention) or an absolute path. I do not know how to get the path to the folder containing the config file, so I pass it in from the outside. The code will crash if users do not set configDir after updating, or likely to crash if the `name` becomes `configDir`.

```
export function includeIgnoreFile(ignoreFilePath, configDir, name) {
	if (!path.isAbsolute(ignoreFilePath)) {
		ignoreFilePath = path.join(configDir, ignoreFilePath)
	}

	const ignoreFile = fs.readFileSync(ignoreFilePath, "utf8");
	const lines = ignoreFile.split(/\r?\n/u);
	
	let dir = path.relative(configDir, path.dirname(ignoreFilePath));
	if (process.platform == 'win32') {
		dir = dir.replace(/\\/g, '/');
	}

	return {
		name: name || "Imported .gitignore patterns",
		ignores: lines
			.map(line => line.trim())
			.filter(line => line && !line.startsWith("#"))
			.map(line => {
				const isNegated = line.startsWith("!");
				const negatedPrefix = isNegated ? "!" : "";
				const pattern = isNegated ? line.slice(1) : line;
				return negatedPrefix + path.posix.join(dir, convertIgnorePatternToMinimatch(pattern));
			}),
	};
}

// usage
includeIgnoreFile('folder/.gitignore', import.meta.dirname);
```

### Participation

- [ ] I am willing to submit a pull request for this change.

### Additional comments

_No response_

kirkwaiblinger:
#### Prerequisites checklist

#### AI acknowledgment

    - (for the most part; some AI use for the added documentation)
- [x] (If the above is not checked) I have reviewed the AI-generated content before submitting.

fixes #329 

Deprecate `includeIgnoreFile()`,`convertIgnorePatternToMinimatch()` in @eslint/compat. Add and export new `includeIgnoreFile()` (documented),`convertIgnorePatternToMinimatch()` (undocumented) in @eslint/config-helpers.

The new `convertIgnorePatternToMinimatch()` is unaltered.
The new `includeIgnoreFile()`:
  - adds a `gitignoreResolution` option which can be `false` (default; interprets patterns relative to config file) or `true` (interprets patterns relative to ignore file).
  - can accept an array of ignore file paths rather than only a single ignore file path.

#### Related Issues

#329 

---

There is a companion PR in eslint repo: https://github.com/eslint/eslint/pull/20735


packages/config-helpers/tests/ignore-file.test.js
```diff
@@ -0,0 +1,184 @@
...
+			const ignoreFilePath = [
+				"../tests/fixtures/ignore-files/gitignore1.txt",
+			];
+			assert.throws(() => {
+				includeIgnoreFile(ignoreFilePath);
+			}, /The ignore file location must be an absolute path. Received .*/u);
+		});
+
+		it("should return an object with an `ignores` property", () => {
+			const ignoreFilePath = fileURLToPath(
```
DMartens, line 73: I think we should simplify this and in other spots by using [import.meta.resolve](https://nodejs.org/docs/latest/api/esm.html#importmetaresolvespecifier) (requires Node v18+)
```suggestion
			const ignoreFilePath = import.meta.resolve("../tests/fixtures/ignore-files/gitignore1.txt")
```

packages/config-helpers/tests/ignore-file.test.js
```diff
@@ -0,0 +1,184 @@
...
+			});
+		});
+	});
+});
+
+// These tests ensure that the `includeIgnoreFile` is a superset of the
+// functionality of the `includeIgnoreFile` in `@eslint/compat`. The only
+// discrepancy is the default name has been changed to reflect that it is
+// really eslintignore mode by default.
+describe("`includeIgnoreFile` compat with @eslint/compat", () => {
```
DMartens, line 176: The tests for the object form options (`{ mode: string, name: string }`) seem to be missing?

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,192 @@
...
+			);
+		}
+		if (!path.isAbsolute(ignorePath)) {
+			throw new Error(
+				`The ignore file location must be an absolute path. Received ${ignorePath}`,
+			);
+		}
+	}
+
+	const { mode, name } = (() => {
```
DMartens: I think this should be split up into two functions as it sets both the defaults and then validates them.

packages/config-helpers/tests/ignore-file.test.js
kirkwaiblinger: This section of tests was copied verbatim from the `@eslint/compat` repo (with the one noted change). That version of `includeIgnoreFile` only supports a string for its second argument.

packages/config-helpers/tests/ignore-file.test.js
kirkwaiblinger: Huh... I never knew about `import.meta.resolve`. But in any case it looks like that [still gets us a URL](https://nodejs.org/docs/latest/api/esm.html#importmetaresolvespecifier), which would need to be converted to a path with `fileURLToPath()`. So I'm not sure that that gains us anything? Unless I'm misunderstanding.

At the end of the day, this is really just the workaround for not having access to `import.meta.dirname`, which the linter complains about (due to supporting older versions of node)

packages/config-helpers/tests/ignore-file.test.js
kirkwaiblinger: Oh, I see, though, it looks like I haven't added a test with that combination in the preceding section. I'll go ahead and add one 👍

packages/config-helpers/src/ignore-file.js
kirkwaiblinger: Changed 👍

packages/compat/README.md
```diff
@@ -149,6 +149,9 @@ module.exports = defineConfig([
 
 ### Including Ignore Files
 
+> [!WARNING]
+> The `includeIgnoreFile()` exported by this package has been deprecated ([eslint/rewrite#329](https://github.com/eslint/rewrite/issues/329)). Use the `convertIgnorePatternToMinimatch()` function exported by `@eslint/config-helpers` instead (also available at `eslint/config`). This section is only preserved for historical reasons.
```
mdjermanovic:
```suggestion
> The `includeIgnoreFile()` exported by this package has been deprecated ([eslint/rewrite#329](https://github.com/eslint/rewrite/issues/329)). Use the `includeIgnoreFile()` function exported by `@eslint/config-helpers` instead (also available at `eslint/config`). This section is only preserved for historical reasons.
```

It would make more sense to use the direct replacement?

kirkwaiblinger: Oh yeah, that's a typo 😄

packages/compat/tests/ignore-file.test.js
```diff
@@ -1,5 +1,6 @@
 /**
- * @filedescription Fixup tests
+ * @filedescription Tests for `includeIgnoreFile()` and `convertIgnorePatternToMinimatch()`
```
mdjermanovic:
```suggestion
 * @fileoverview Tests for `includeIgnoreFile()` and `convertIgnorePatternToMinimatch()`
```

Since we're already updating this, we can also fix the tag to `@fileoverview` which we normally use (I'm not sure if `@filedescription` is actually valid).

packages/compat/README.md
```diff
@@ -149,6 +149,9 @@ module.exports = defineConfig([
 
 ### Including Ignore Files
 
+> [!WARNING]
+> The `includeIgnoreFile()` exported by this package has been deprecated ([eslint/rewrite#329](https://github.com/eslint/rewrite/issues/329)). Use the `includeIgnoreFile()` function exported by `@eslint/config-helpers` instead (also available at `eslint/config`). This section is only preserved for historical reasons.
```
mdjermanovic: Perhaps it would be better to use just standard markdown in README files (e.g., ``**Deprecated**: The `includeIgnoreFile() ...``). I'm not sure if GitHub alerts would be rendered as expected on npm, JSR, and other places that show published md files.

packages/config-helpers/README.md
```diff
@@ -74,6 +74,84 @@ export default defineConfig([
...
+		],
+		{ mode: "gitignore" },
+	),
+	// ...
+);
+```
+
+### `convertIgnorePatternToMinimatch()`
+
+This is used under the hood by `includeIgnoreFile()` to convert patterns found in eslintignore/gitignore files. You can use it to construct your own ignore objects if you wish:
```
mdjermanovic: `convertIgnorePatternToMinimatch` was purposely undocumented (see https://github.com/eslint/rewrite/pull/47) as it was intended for internal use only by the `@eslint/migrate-config` package, so I'm not sure if we want to document it now.

kirkwaiblinger: Ah - I wasn't aware of that context. I'm happy to do whatever the team prefers here. I guess the levers are:

1. Reexport `convertIgnorePatternToMinimatch()` from `eslint/config`?
2. Export `convertIgnorePatternToMinimatch()` from `@eslint/config-helpers` main entry point?
3. Export `convertIgnorePatternToMinimatch()` from `@eslint/config-helpers` at a use-at-your-own-risk entry opint?
4. (If exported in any way) Document `convertIgnorePatternToMinimatch()`?

I guess we could always start with: don't document and don't export the new copy of `convertIgnorePatternToMinimatch()` in any way. If a use case arises an appropriate export can be added.

packages/compat/tests/ignore-file.test.js
kirkwaiblinger: Sounds good. Changed 👍

packages/compat/README.md
kirkwaiblinger: touche. Changed 👍

packages/config-helpers/README.md
fasttime: > `convertIgnorePatternToMinimatch` was purposely undocumented (see #47) as it was intended for internal use only by the `@eslint/migrate-config` package, so I'm not sure if we want to document it now.

In that case, maybe it's not necessary to export `convertIgnorePatternToMinimatch` from `config-helpers`?

mdjermanovic: `@eslint/migrate-config` [uses](https://github.com/eslint/rewrite/blob/cb7e47e77389724d02510d61a636d8a96d1cb31c/packages/migrate-config/src/migrate-config.js#L16) `convertIgnorePatternToMinimatch`, currently from `@eslint/compat`. If we're going to deprecate it in `@eslint/compat`, parhaps it would make sense to switch `@eslint/migrate-config` to using it from `@eslint/config-helpers`?

fasttime: Okay, then I'd say let's switch `@eslint/migrate-config` to depending on `@eslint/config-helpers`, export `convertIgnorePatternToMinimatch` from `@eslint/config-helpers`, and deprecate `convertIgnorePatternToMinimatch` in `@eslint/compat`.

kirkwaiblinger: Ok I updated the eslint PR (https://github.com/eslint/eslint/pull/20735) to not reexport `convertIgnorePatternToMinimatch` at `eslint/config` and to not document it.

I updated this PR to remove the documentation for `convertIgnorePatternToMinimatch` as well.

I tried to make the change to migrate-config, but I'm not sure whether it can be done in this PR since `@eslint/migrate-config` will need to reference the next released version of `@eslint/config-helpers`. In any case, the change will be as trivial as 

```diff
diff --git a/packages/migrate-config/package.json b/packages/migrate-config/package.json
index 4a308cf06..80d69ea8f 100644
--- a/packages/migrate-config/package.json
+++ b/packages/migrate-config/package.json
@@ -47,7 +47,7 @@
     "node": "^20.19.0 || ^22.13.0 || >=24"
   },
   "dependencies": {
-    "@eslint/compat": "^2.0.5",
+    "@eslint/config-helpers": "^0.6.0",
     "@eslint/eslintrc": "^3.3.5",
     "camelcase": "^8.0.0",
     "espree": "^10.4.0",
diff --git a/packages/migrate-config/src/migrate-config.js b/packages/migrate-config/src/migrate-config.js
index b1c33fd7b..acc3d55ce 100644
--- a/packages/migrate-config/src/migrate-config.js
+++ b/packages/migrate-config/src/migrate-config.js
@@ -13,7 +13,7 @@ import { Legacy } from "@eslint/eslintrc";
 import camelCase from "camelcase";
 import pluginsNeedingCompat from "./compat-plugins.js";
 import configsNeedingCompat from "./compat-configs.js";
-import { convertIgnorePatternToMinimatch } from "@eslint/compat";
+import { convertIgnorePatternToMinimatch } from "@eslint/config-helpers";
 import * as espree from "espree";
```

(assuming 0.6.0 is the next release for config-helpers)

Also - I'm thinking maybe `migrate-config` should be updated to use the new version of `includeIgnoreFile` in its codegen (rather than the now-deprecated version in `@eslint/compat`) but this should probably also wait until the config-helpers package is actuallly released.

packages/migrate-config/package.json
```diff
@@ -47,7 +47,7 @@
     "node": "^20.19.0 || ^22.13.0 || >=24"
   },
   "dependencies": {
-    "@eslint/compat": "^2.0.5",
+    "@eslint/config-helpers": "^0.5.5",
```
kirkwaiblinger, line 49: As explained in more detail in https://github.com/eslint/rewrite/pull/430#discussion_r3066050332 this will not actually work with config-helpers 0.5.5, so I'm not sure what to do here; it needs to use the release of config-helpers that contains the changes in this PR.

packages/config-helpers/README.md
```diff
@@ -74,6 +74,84 @@ export default defineConfig([
...
+			path.join(import.meta.dirname, ".gitignore"),
+			path.join(import.meta.dirname, "packages/lib/.gitignore"),
+		],
+		{ mode: "gitignore" },
+	),
+	// ...
+);
+```
+
+### `convertIgnorePatternToMinimatch()`
```
mdjermanovic: We can remove this from the docs here.

packages/migrate-config/package.json
mdjermanovic: This is okay, our release process will handle it. Release commit will update this to 0.6.0, and `@eslint/config-helpers` v0.6.0 will be released before the new version of `@eslint/migrate-config`.

packages/config-helpers/README.md
mdjermanovic: > Also - I'm thinking maybe `migrate-config` should be updated to use the new version of `includeIgnoreFile` in its codegen (rather than the now-deprecated version in `@eslint/compat`) but this should probably also wait until the config-helpers package is actuallly released.

We could consider updating it to use `includeIgnoreFile` from `eslint/config` after releasing `eslint` with changes from https://github.com/eslint/eslint/pull/20735.

packages/config-helpers/README.md
```diff
@@ -74,6 +74,84 @@ export default defineConfig([
 ]);
 ```
 
+### `includeIgnoreFile()`
+
+The `includeIgnoreFile()` function reads an ignore file (such as a `.gitignore`) and returns a config object with the patterns converted to a global ignores object. Pass the absolute path to the ignore file as the first argument:
```
mdjermanovic:
```suggestion
The `includeIgnoreFile()` function reads a file with gitignore-style patterns (such as a `.gitignore`) and returns a config object with the patterns converted to a global ignores object. Pass the absolute path to the ignore file as the first argument:
```

packages/config-helpers/README.md
```diff
@@ -74,6 +74,84 @@ export default defineConfig([
...
+
+```js
+// eslint.config.js
+
+import { includeIgnoreFile } from "@eslint/config-helpers";
```
mdjermanovic:
```suggestion
import { defineConfig, includeIgnoreFile } from "@eslint/config-helpers";
```

packages/config-helpers/README.md
```diff
@@ -74,6 +74,84 @@ export default defineConfig([
...
+	// ...
+);
+```
+
+#### Options
+
+The second argument is an optional options object:
+
+- **`mode`** : Controls how ignore patterns are interpreted.
+    - `"eslintignore"` (default) — patterns are resolved relative to the current working directory, matching the behavior of `.eslintignore` files.
```
mdjermanovic:
```suggestion
    - `"eslintignore"` (default) — patterns are resolved relative to the location of the configuration file.
```

I believe this would be a more correct explanation. Also, I think there's no need to mention legacy .eslintignore files in config-helpers.

kirkwaiblinger: Ah - yes, you're right about the CWD vs config file thing, changing that 👍 

On mentioning "eslintignore": Shall we also change the name of the mode then? The reason that I had worded it the way it was was partially to make the default value of `mode: 'eslintignore'` seem like a plausible choice. As a user encountering this as a new API I think I would otherwise be perplexed/annoyed that `mode: "gitignore"` isn't the default. Do you have thoughts as to how to present the default behavior in a way that implies a motivation for it?

kirkwaiblinger: Especially on the companion PR (https://github.com/eslint/eslint/pull/20735), I think this will definitely require some thought... currently I have this: https://github.com/eslint/eslint/blob/a25e95f4bec496c129a01a41eb95da024d113cd9/docs/src/use/configure/ignore.md#L220-L307

mdjermanovic: As for the reasoning for the default mode, since users would typically want the other mode, I think it would be best to say that the default mode is the mode it is because it matches the behavior of the first version of this helper, i.e., no particularly good reason other than backwards compatibility. Also, eslint config file and .gitinore file are usually in the same place (project root) so in that case there is no difference between modes.

As for the modes/option name, maybe a boolean "patterns match relative to the location of the ignore file" would be clearer, just not sure how to put this into an option name that wouldn't be overly verbose.

mdjermanovic:
The `lint:unused` check is failing now:

```
> eslint-rewrite@1.0.0 lint:unused
> knip

Unused devDependencies (1)
eslint  packages/migrate-config/package.json:44:6
```

I think we can just remove `eslint` from devDependencies in packages/migrate-config/package.json to fix this.

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	const mode = options?.mode ?? "eslintignore";
+	if (!(mode === "gitignore" || mode === "eslintignore")) {
+		throw new Error(
+			'The `mode` option must be specified as "gitignore" or "eslintignore"',
+		);
+	}
+
+	const name = options?.name ?? `Imported .${mode} patterns`;
```
mdjermanovic:
```suggestion
	const name = options?.name ?? "Imported .gitignore patterns";
```

I think this should always be ".gitignore" as it refers to .gitignore-style patterns rather than the filename or the mode.

packages/config-helpers/README.md
mdjermanovic: Maybe `matchRelativeToIgnoreFile: true`?

By default (`false`), patterns match relative to the location of the configuration file in which they are included, or relative to CWD when the configuration file is used with the `--config` option.

When `true`, patterns always match relative to the location of the ignore file.

fasttime: If we want to use a boolean option, I think it would be better to refer to the usage of the file as a `.gitignore` (maybe `gitignoreResolution`?). This would make it more obvious for user to understand what value to use, instead of focusing on the different base location of relative paths in `.eslintignore` files, which is a poorly documented caveat of the specification.

mdjermanovic: `gitignoreResolution` sounds good to me.

kirkwaiblinger: Ok, updated to `gitignoreResolution: boolean`

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for others to verify.

packages/compat/src/ignore-file.js
```diff
@@ -24,6 +24,9 @@ import path from "node:path";
  * Converts an ESLint ignore pattern to a minimatch pattern.
  * @param {string} pattern The .eslintignore or .gitignore pattern to convert.
  * @returns {string} The converted pattern.
+ *
+ * @deprecated Use the `convertIgnorePatternToMinimatch()` function exported by
+ * `@eslint/config-helpers` instead (also available at `eslint/config`).
```
fasttime: Just not sure if `convertIgnorePatternToMinimatch` should be exported from `"eslint/config"` as well. What is the use case for that?

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Reads an ignore file(s) and returns an object(s) with the ignore patterns.
+ *
+ * @param {string[] | string} ignoreFilePathArg The path(s) of the ignore file(s) to include.
+ * @param {object} [options]
+ * @param {boolean} [options.gitignoreResolution] Whether to interpret the contents of the ignore file relative to the config file or the ignore file.
+ * - gitignoreResolution: false (default): Interprets the ignore patterns relative to the config file
+ * - gitignoreResolution: true: Interprets the ignore patterns relative to the ignore file
+ *
+ * @param {string} [options.name] The name to give the output config object(s).
```
fasttime: `options` could be a `string` as well?

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * @overload
+ *
+ * Reads ignore files and returns objects with the ignore patterns.
+ *
+ * @param {string[]} ignoreFilePathArg The paths of ignore files to include
+ * @param {object} [options]
+ * @param {boolean} [options.gitignoreResolution] Whether to interpret the contents of the ignore file relative to the config file or the ignore file.
```
fasttime: Since this is the overload for the array form of `ignoreFilePathArg`

```suggestion
 * @param {boolean} [options.gitignoreResolution] Whether to interpret the contents of the ignore files relative to the config file or the ignore file.
```

fasttime: Or maybe we could add a typedef for the options type to avoid repeating JSDoc for `gitignoreResolution` and `name`?

packages/compat/src/ignore-file.js
kirkwaiblinger: that comment was left over by accident from a previous version of this PR; removed 👍

packages/config-helpers/src/ignore-file.js
kirkwaiblinger: Good call on the typedef. I've made changes in this area, lmk what you think

packages/config-helpers/src/ignore-file.js
kirkwaiblinger: Yeah... I didn't initially add that in the typing because I didn't really want to advertise it; promoting the object form will nudge people towards using the `gitignoreResolution: true`, which is a good thing IMO. 

But I've gone ahead and updated this to acknowledge the `string` form too, and included a note in the JSDoc

packages/config-helpers/tests/types/types.test.ts
```diff
@@ -229,3 +230,43 @@ globalIgnores([1]);
...
+	gitignoreResolution: true,
+	name: "falafel",
+})
+	// @ts-expect-error -- return type shouldn't be able to access array method
+	.map(config => config.ignores);
+
+// should be able to provide a string options argument for compatibility reasons.
+includeIgnoreFile("foo", "string-name");
+
+// @ts-expect-error -- options should not a number.
```
fasttime:
```suggestion
// @ts-expect-error -- options should not be a number.
```

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,204 @@
...
+
+	return { gitignoreResolution, name };
+}
+
+/**
+ * @overload
+ *
+ * Reads ignore files and returns objects with the ignore patterns.
+ *
+ * @param {string[]} ignoreFilePathArg The paths of ignore files to include
```
fasttime:
```suggestion
 * @param {string[]} ignoreFilePathArg The paths of ignore files to include.
```

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,204 @@
...
+ * @param {IncludeIgnoreFileOptions} [options]
+ * @returns {Config[]}
+ */
+
+/**
+ * @overload
+ *
+ * Reads an ignore file and returns an object with the ignore patterns.
+ *
+ * @param {string} ignoreFilePathArg The path of the ignore file to include
```
fasttime:
```suggestion
 * @param {string} ignoreFilePathArg The path of the ignore file to include.
```

packages/config-helpers/src/ignore-file.js
fasttime: That seems fine now, thanks.

fasttime (commented):
Just some minor typos.

packages/compat/README.md
lumirlumir, line 1: I think it would be nice to include the `deprecated` information in the very first section as well.

https://github.com/eslint/rewrite/blob/331cf8dda17ffab3101a3f4f8bfff595de5f02db/packages/compat/README.md?plain=1#L36

packages/config-helpers/README.md
lumirlumir, line 1: As I recall, there was a mention that it might be a good opportunity to document `convertIgnorePatternToMinimatch()`; although it isn’t exported from `eslint/config`, it’s available in `@eslint/config-helpers`.

https://github.com/eslint/rewrite/issues/329#issuecomment-4186746147

If possible, could we also add documentation for `convertIgnorePatternToMinimatch()`?

packages/config-helpers/tests/types/types.test.ts
lumirlumir, line 1: I think it would be nice to add type tests for `convertIgnorePatternToMinimatch()` :)

lumirlumir (changes requested):

packages/config-helpers/README.md
mdjermanovic: It's intentionally undocumented (https://github.com/eslint/rewrite/pull/430#discussion_r3058800836).

packages/config-helpers/tests/types/types.test.ts
kirkwaiblinger: added 👍

packages/compat/README.md
kirkwaiblinger: added 👍

packages/config-helpers/src/index.js
```diff
@@ -4,3 +4,7 @@
 
 export { defineConfig } from "./define-config.js";
 export { globalIgnores } from "./global-ignores.js";
+export {
+	includeIgnoreFile,
+	convertIgnorePatternToMinimatch,
+} from "./ignore-file.js";
```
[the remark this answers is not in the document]
mdjermanovic: Addressed in https://github.com/eslint/rewrite/pull/430/commits/c3a24c2186c8586ca72a02a366096b3ae5c05ef7.

packages/config-helpers/tests/types/types.test.ts
```diff
@@ -229,3 +230,43 @@ globalIgnores([1]);
...
+
+includeIgnoreFile(".gitignore").ignores;
+includeIgnoreFile(".gitignore", {}).ignores;
+includeIgnoreFile(".gitignore", { gitignoreResolution: true, name: "falafel" })
+	.ignores;
+
+// array of string paths should return an array of config objects
+includeIgnoreFile([".gitignore", ".eslintignore"]).map(
+	config => config.ignores,
+);
```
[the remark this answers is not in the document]
mdjermanovic: Addressed in https://github.com/eslint/rewrite/pull/430/commits/a49ef42a9389c1c5904df9067b740d95c34a3067

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,204 @@
...
+ * - gitignoreResolution: true: Interprets the ignore patterns in a file relative to the ignore file
+ * @property {string} [name] The name to give the output config object(s).
+ */
+
+/**
+ * Options for `includeIgnoreFile()`. May be provided as an object or, for
+ * legacy compatibility with `@eslint/compat`, as a string which is treated as
+ * the `name` option.
+ * @typedef {IncludeIgnoreFileOptionsObject | string} IncludeIgnoreFileOptions
+ */
```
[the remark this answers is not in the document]
mdjermanovic: PR description has been updated accordingly.

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,204 @@
...
+	if (typeof options === "string") {
+		return { gitignoreResolution: false, name: options };
+	}
+
+	const gitignoreResolution = options?.gitignoreResolution ?? false;
+	if (typeof gitignoreResolution !== "boolean") {
+		throw new Error(
+			"The `gitignoreResolution` option must be specified a boolean or omitted",
+		);
+	}
```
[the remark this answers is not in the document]
mdjermanovic: Addressed in https://github.com/eslint/rewrite/pull/430/commits/e2fd6b476e46e3d6070d4b881e2b22af8c569408.

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for others to verify their review comments.

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,211 @@
...
+	const gitignoreResolution = optionsObject.gitignoreResolution ?? false;
+	if (typeof gitignoreResolution !== "boolean") {
+		throw new Error(
+			"The `gitignoreResolution` option must be specified a boolean or omitted",
+		);
+	}
+
+	const name = optionsObject.name ?? `Imported .gitignore patterns`;
+	if (typeof name !== "string") {
+		throw new Error(
```
lumirlumir:
```suggestion
		throw new TypeError(
			"The options argument to `includeIgnoreFile()` should be an object or a string.",
		);
	}

	const gitignoreResolution = optionsObject.gitignoreResolution ?? false;
	if (typeof gitignoreResolution !== "boolean") {
		throw new TypeError(
			"The `gitignoreResolution` option must be specified a boolean or omitted",
		);
	}

	const name = optionsObject.name ?? `Imported .gitignore patterns`;
	if (typeof name !== "string") {
		throw new TypeError(
```

Maybe `TypeError` would be more specific here? When checking an argument’s type, `TypeError` seems more precise, and it looks like other places follow the `TypeError` approach for argument validation as well.

https://github.com/eslint/rewrite/blob/c872ca86421793467c52e3003f018f34d74c48d3/packages/config-helpers/src/global-ignores.js#L30-L36

---

It’s an optional suggestion, but I think adding a `TypeError` check to the tests would be a nice-to-have here.

```js
assert.throws(
	() => includeIgnoreFile(123),
	{
		name: "TypeError",
		message: /error message/,
	},
);
```

packages/config-helpers/src/ignore-file.js
```diff
@@ -0,0 +1,211 @@
...
+		? ignoreFilePathArg
+		: [ignoreFilePathArg];
+	for (const ignorePath of ignoreFilePaths) {
+		if (typeof ignorePath !== "string") {
+			throw new Error(
+				"The first argument to `includeIgnoreFile()` should be a string or array of strings",
+			);
+		}
+		if (!path.isAbsolute(ignorePath)) {
+			throw new Error(
```
lumirlumir:
```suggestion
			throw new TypeError(
				"The first argument to `includeIgnoreFile()` should be a string or array of strings",
			);
		}
		if (!path.isAbsolute(ignorePath)) {
			throw new TypeError(
```

Same here :)

packages/config-helpers/tests/ignore-file.test.js
```diff
@@ -0,0 +1,244 @@
...
+			});
+		});
+
+		it("should return an object with a custom name", () => {
+			const ignoreFilePath = fileURLToPath(
+				new URL(
+					"../tests/fixtures/ignore-files/gitignore1.txt",
+					import.meta.url,
+				),
+			);
```
lumirlumir: Would it make sense to extract `ignoreFilePath` into a top-level helper variable here?

It looks like `ignoreFilePath` is used six times across the tests, so I think moving it could make the code a bit more concise.

packages/config-helpers/tests/ignore-file.test.js
```diff
@@ -0,0 +1,244 @@
...
+		});
+
+		it("should handle when both name and gitignoreResolution are specified", () => {
+			const ignoreFilePath = fileURLToPath(
+				new URL(
+					"../tests/fixtures/ignore-files/gitignore1.txt",
+					import.meta.url,
+				),
+			);
+			const basePath = fileURLToPath(
```
lumirlumir: Moving `basePath` to the top level could also be a nice option, since it’s used twice (though I’m not strongly opinionated about it).

lumirlumir (changes requested):

packages/config-helpers/README.md
```diff
@@ -74,6 +74,59 @@ export default defineConfig([
...
+	}),
+	// ...
+]);
+```
+
+#### Options
+
+The second argument is an optional options object:
+
+- **`gitignoreResolution`** (`boolean`) : Controls how ignore patterns are interpreted.
```
lumirlumir:
```suggestion
- **`gitignoreResolution`** (`boolean`): Controls how ignore patterns are interpreted.
```

Non-blocking, but there's a small typo here.

lumirlumir (dismissed):
LGTM. Thanks for your time and effort on this PR!

Would like @mdjermanovic to verify the changes before merging.

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for @fasttime and @DMartens to verify.

fasttime (dismissed):
LGTM.

packages/config-helpers/tests/ignore-file.test.js
```diff
@@ -0,0 +1,239 @@
...
+						"**/.cache/",
+						".vuepress/dist",
+						"*/foo.js",
+						"dir/**/*",
+					],
+				},
+			]);
+		});
+
+		// convert above to for ... of loop
```
DMartens: I think this is an unresolved TODO comment related to line 61 which uses `Array#forEach` to create a parameterised test case.

DMartens:
Changes LGTM, thank you for the continued work on this.
Only found a seemingly unresolved TODO comment.

packages/config-helpers/tests/ignore-file.test.js
kirkwaiblinger: Gah! that was a prompt to copilot that I forgot to remove, yep :face_with_peeking_eye:. Thanks for pointing that out.

DMartens (approved):

PR #445, opened 2026-05-03
fix: add missing .d.cts files while building plugin-kit

liangmiQwQ:
#### Prerequisites checklist

#### AI acknowledgment

Please read [sxzz/rolldown-plugin-dts#1230](https://github.com/sxzz/rolldown-plugin-dts/issues/232#issuecomment-4364917941). ESLint should emit .d.ts for commonjs for downstream's build.

Copied `core` package's *bundling* logic

#### Related Issues

[sxzz/rolldown-plugin-dts#1230](https://github.com/sxzz/rolldown-plugin-dts/issues/232#issuecomment-4364917941)


liangmiQwQ:
bro... I have to say it's really ...

image

DMartens:
I cannot reproduce this problem as `dist/cjs/index.d.cts` for `plugin-kit` is created.
You can see this in the package.json script `build:cts` you want to modify.
Regarding the EasyCLA problem, please retry maybe in a different browser, as we cannot proceed otherwise.

liangmiQwQ:
> I cannot reproduce this problem as `dist/cjs/index.d.cts` for `plugin-kit` is created. You can see this in the package.json script `build:cts` you want to modify. Regarding the EasyCLA problem, please retry maybe in a different browser, as we cannot proceed otherwise.

I want to point out the file we need is the `type.d.ts`, which is missing in `cjs` directory. I think we should just do the things ESlint has already done in `core` package instead of modifying or calling the global building script (as there are no other imports in `type.cts`).

https://github.com/eslint/rewrite/blob/70b69970a2e0d9bdba8c2e9c19beb6f5b35d5db0/packages/core/package.json#L20

~~And I will try to solve EasyCLA asap.~~ Thank you.

Update: EasyCLA was solved.

lumirlumir (commented):
Actually, it’s a bit difficult to pinpoint the root cause here, since your reproduction link is about the rolldown integration.

Could you please share a failing minimal reproduction link without the rolldown integration?

Also, updating the PR template to use our bug template would be helpful here, so others can more easily recognize the real problem.

liangmiQwQ:
> Could you please share a failing minimal reproduction link without the rolldown integration?

Umm... this might be a little bit hard to describe or be shown in a reproduction repo, I think I need to re-explain the question.

`@eslint/plugin-kit` exposes `index.d.(c)ts` for both esm and cjs, which imports `types.(c)ts`. The bundler will automatically redirects all pure `.(c)ts` files to `.d.(c)ts` as a type declaration (most of the bundlers do it, like rollup, as well as rolldown).

And for the current `@eslint/plugin-kit`'s building process, the `types.d.ts` file only exists under `esm` directory, not `cjs` directory, which causes bundler can't find the target `.d.cts` when using CommonJS.

So, for packages who depend on and bundle `@eslint/plugin-kit`.
- If they are using `Rollup`, it will automatically generate a `types.d.cts` as it is missing.
- If they are using rolldown (like above), it won't generate this file, make bundling error happens.

I've talked with rolldown team, they thought it is expected and it's an ESLint's side building config issue. It makes sense, because the `esm` side has `types.d.ts` (Generated by `tsc`), but the `cjs` has no `types.d.cts`, which is missing exactly.

Hope it's helpful, thanks.

DMartens:
Thank you for the explanation.
I think that this is not bundler-specific problem and agree that this should be fixed on our side.
The changes make sense to me, leaving open for @lumirlumir.

DMartens (approved):

liangmiQwQ:
Thank You!

liangmiQwQ:
Hello @lumirlumir, sorry for pinging, but it seems we need to make it active!

lumirlumir:
@liangmiQwQ 

Apologies for the delay. I've been a bit busy lately, and I'll make sure to review this change by tomorrow. Thanks in advance for your patience!

liangmiQwQ:
Never mind!

However, I'm not sure whether the CI falling is related to my changes in 501d4ad.

lumirlumir (approved):
LGTM, thanks for the time you spent waiting.

I like this change, and I’m leaving some notes here for future reference.

After looking into this a bit more, I don’t think the missing `dist/cjs/types.d.cts` is a problem for standard TypeScript type resolution by itself. `dist/cjs/types.cts` is not exposed as a public API entry point and is only referenced from `index.d.cts`, which TypeScript can resolve correctly.

So I understand this change mainly as a packaging compatibility improvement for downstream declaration-bundling tools, such as `rolldown`/`rolldown-plugin-dts`, which may not re-emit declarations for files inside `node_modules`. Shipping the CJS declaration graph in a complete `.d.cts` form makes the package more robust for those integrations and will not affect ordinary users.

lumirlumir:
> However, I'm not sure whether the CI falling is related to my changes in https://github.com/eslint/rewrite/commit/501d4ad45e1a12e63a56d8b0a80d24cbe4fe6c3f.

I think that was a temporary issue from a downstream dependency and was fixed in a separate PR, so no worries at all!

PR #452, opened 2026-05-15
ci: declare workflow-level `contents: read` on 4 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 4 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/ci-bun.yml
+++ b/.github/workflows/ci-bun.yml
@@ -11,6 +11,9 @@ on:
             - main
             - "release/*/v**"
 
+permissions:
+    contents: read
+
 jobs:
     ci-bun:
         uses: eslint/workflows/.github/workflows/ci-bun.yml@main
--- a/.github/workflows/ci-package-manager.yml
+++ b/.github/workflows/ci-package-manager.yml
@@ -9,6 +9,9 @@ on:
         branches:
             - main
 
+permissions:
+    contents: read
+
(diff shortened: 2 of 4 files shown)


lumirlumir (approved):
LGTM, thanks!

lumirlumir (changes requested):
Ah, I missed the CI check. The changes look good to me, but could you take a look at the formatting failure? The "Are the types wrong" CI failure should be resolved once the branch is rebased.

arpitjain099:
Thanks @lumirlumir, good catch. Two fixes pushed:

1. Re-indented the `permissions:` block from 2 to 4 spaces to match the rest of the workflow YAML (that was the Verify Files failure).
2. Rebased on `main` so the "Are the types wrong" check resolves.

Let me know if you spot anything else.

lumirlumir (approved):
LGTM, thanks!

PR #457, opened 2026-05-16
feat!: migrate migrate-config to new includeIgnoreFile

kirkwaiblinger:
#### Prerequisites checklist

#### AI acknowledgment

migrate-config will now generate a config that uses the new `includeIgnoreFile` from `eslint/config` rather than the now-deprecated version from `@eslint/compat`.

#### Related Issues

Followup to https://github.com/eslint/rewrite/pull/430#discussion_r3067961400


jookira:
This change would break the migration path for anyone landing on ESLint v9, as `includeIgnoreFile` was only added to `eslint/config` in v10.4.0.

It’s probably time to consider freezing this package altogether.

Edit: I just saw the discussion about the new codemod package in https://github.com/eslint/eslint/issues/20442 and proposed the transform there.

lumirlumir:
I think the CI failure related to “Are the types wrong” will be resolved once this branch is rebased.

kirkwaiblinger:
I am entirely happy for this to be abandoned but would just like for an ESLint team member to make the call either way since I also don't want to unilaterally abandon a followup which a team member was part of.

lumirlumir (commented):
@kirkwaiblinger I’m sorry for the delay.

I’ve read through all the threads in https://github.com/eslint/eslint/issues/20442, and I think the team is currently investigating the possibility of consolidating `@eslint/migrate-config` and `eslint-transforms` into Codemod.

While I agree that this change would be helpful for ESLint v10 migration support, which may happen in the next major release of `@eslint/migrate-config`, there’s currently no definite consensus on how we’ll treat the `@eslint/migrate-config` package in the future.

For this reason, I’d like to raise this question to gather consensus. Thanks for your patience in advance while we work toward a decision!

---

To summarize the question for the team: @eslint/eslint-team 

What is the future of `@eslint/migrate-config` if we’re investigating the possibility of consolidating it into Codemod?

Should package updates be frozen until a decision is made, or is it okay to look into the possibility of ESLint v10 support including this PR?

lumirlumir:
**TSC Summary**: This PR is mainly about the future of the `@eslint/migrate-config` package, since `@eslint/v8-to-v9-config`, which has been verified in https://github.com/eslint/eslint/issues/20442 and is ready for a blog announcement, could replace it.
This PR also changes the implementation to use `includeIgnoreFile` from the `eslint/config` subpath import. Since this feature was added in ESLint v10.4.0, it would break the existing v8-to-v9 migration tooling. This change need to be added in the next major version of `@eslint/migrate-config` instead, so it doesn’t break the current migration solution.

**TSC Question**: Given that the v8-to-v9 and v9-to-v10 codemods could cover the `migrate-config` package, should this package be deprecated and frozen for new feature requests, with all codemods unified in the codemods repo moving forward? Or do we plan to maintain both packages?

lumirlumir (commented):
In the TSC meeting on June 11, 2026, it was determined that we should move forward in this way:

- We don’t want to freeze or deprecate `migrate-config` right now
- We’d like to maintain v9 output behind a `--target-version` option, with 10 as the default and 9 as an option
- This PR is a breaking change

@kirkwaiblinger, are you still available to move forward with the approach above? It would be great if this PR could include the items listed above.

https://discord.com/channels/688543509199716507/688545247843713092/1514726203699757181

I’ve changed the PR title to `feat!` since it’s now a breaking change.

kirkwaiblinger:
> @kirkwaiblinger, are you still available to move forward with the approach above? It would be great if this PR could include the items listed above.

Hi @lumirlumir!

I'm happy to, but could you clarify what actual action items these imply? 

> In the TSC meeting on June 11, 2026, it was determined that we should move forward in this way:
> 
> * We don’t want to freeze or deprecate `migrate-config` right now

This PR does does not freeze/deprecate anything.

> * We’d like to maintain v9 output behind a `--target-version` option, with 10 as the default and 9 as an option

Is it being requested that this be built out as part of this PR? This is the only action item that seems plausible?

> * This PR is a breaking change

I think this just impacts the title (`feat!`) which you've taken care of 🙂

mdjermanovic:
> > * We’d like to maintain v9 output behind a `--target-version` option, with 10 as the default and 9 as an option
> 
> Is it being requested that this be built out as part of this PR? This is the only action item that seems plausible?

Yes, this is the only remaining action item.

packages/migrate-config/README.md
```diff
@@ -55,6 +55,16 @@ npx @eslint/migrate-config .eslintrc.json --gitignore
 bunx @eslint/migrate-config .eslintrc.json --gitignore
 ```
 
+### Targeting a specific ESLint version
+
+By default, this tool targets the latest ESLint version available (v10). You can target v9 by passing the `--targetVersion` option:
```
mdjermanovic: The option should be `--target-version`, not `--targetVersion`. And the valid values are `9` and `10`, not `v9` and `v10`.

packages/migrate-config/src/migrate-config-cli.js
```diff
@@ -25,15 +25,33 @@ import path from "node:path";
...
+			type: "boolean",
+			default: false,
+		},
+		gitignore: {
+			type: "boolean",
+			default: false,
+		},
+		targetVersion: {
+			type: "string",
+			default: "v9",
```
mdjermanovic: Default should be `10`.

packages/migrate-config/src/migrate-config-cli.js
```diff
@@ -25,15 +25,33 @@ import path from "node:path";
...
+		},
+	},
+	allowPositionals: true,
+});
+
+const [configFilePath] = positionals;
+const { commonjs, gitignore, targetVersion } = values;
```
mdjermanovic: `targetVersion` seems to be unused, except for the validation.

packages/migrate-config/tests/fixtures/gitignore-simple-v9/.eslintrc.json
mdjermanovic, line 1: We should add this test to the list to run it:

https://github.com/eslint/rewrite/blob/5ebdeb8832633cda59f47c73db337a186ba8bffe/packages/migrate-config/tests/migrate-config-cli.test.js#L18-L33

packages/migrate-config/src/migrate-config.js
```diff
@@ -395,15 +411,35 @@ function convertGlobPattern(pattern) {
...
+
+		code = `includeIgnoreFile(gitignorePath)`;
 	} else {
-		migration.imports
-			.get("@eslint/compat")
-			.bindings.push("includeIgnoreFile");
+		if (!migration.imports.has("eslint/config")) {
+			migration.imports.set("eslint/config", {
+				bindings: ["includeIgnoreFile"],
+				added: true,
```
mdjermanovic:
```suggestion
```

`eslint/config` is not considered added since it's part of the `eslint` package.

packages/migrate-config/src/migrate-config.js
```diff
@@ -395,15 +411,35 @@ function convertGlobPattern(pattern) {
...
+			});
+		} else {
+			migration.imports
+				.get("eslint/config")
+				.bindings.push("includeIgnoreFile");
+
+			code = `includeIgnoreFile(gitignorePath, { gitignoreResolution: true })`;
+		}
```
mdjermanovic:
```suggestion
		}

		code = `includeIgnoreFile(gitignorePath, { gitignoreResolution: true })`;
```

I believe `code` should be assigned to this value regardless of the condition above.

kirkwaiblinger: Ah - good catch, thanks 👍

packages/migrate-config/src/migrate-config.js
```diff
@@ -78,6 +78,22 @@ class Migration {
 	 */
 	inits = [];
 
+	/**
+	 * The target version for the migration.
+	 * @type {"9"|"10"}
```
lumirlumir:
```suggestion
	 * @type {TargetVersion}
```

Maybe adding a top-level `TargetVersion` internal type and reusing it across the implementation would make sense here? I can see that the `"9" | "10"` pattern is used four times throughout the implementation, and I think introducing a helper type would make it clearer :)

For example:

```js
/** @typedef {"9"|"10"} TargetVersion */
```

packages/migrate-config/src/migrate-config.js
```diff
@@ -78,6 +78,22 @@ class Migration {
...
+
+	constructor(
+		/**
+		 * The target version for the migration.
+		 * @type {"9"|"10"}
+		 */
+		targetVersion,
+	) {
+		this.targetVersion = targetVersion;
+	}
```
lumirlumir:
```suggestion
    /**
	 * Creates a migration object.
	 * @param {"9"|"10"} targetVersion The target version for the migration.
	 */
	constructor(targetVersion) {
		this.targetVersion = targetVersion;
	}
```

Non-blocking stylistic suggestion: since much of the current implementation in `rewrite` consistently uses `@param`-style JSDoc for `constructor` instead of using `@type` directly for parameters, I think using `@param` style would be more consistent across the codebase.

https://github.com/eslint/rewrite/blob/33fc0b32d9ab0965dfe1d2b29d3935057ba8d2ca/packages/config-array/src/config-array.js#L101-L109

https://github.com/eslint/rewrite/blob/33fc0b32d9ab0965dfe1d2b29d3935057ba8d2ca/packages/object-schema/src/object-schema.js#L81-L87

https://github.com/eslint/rewrite/blob/33fc0b32d9ab0965dfe1d2b29d3935057ba8d2ca/packages/object-schema/src/object-schema.js#L94-L100

https://github.com/eslint/rewrite/blob/33fc0b32d9ab0965dfe1d2b29d3935057ba8d2ca/packages/plugin-kit/src/config-comment-parser.js#L71-L77

mdjermanovic (dismissed):
LGTM, thanks! Leaving it open for a second review.

packages/migrate-config/src/migrate-config.js
```diff
@@ -1410,18 +1443,19 @@ function convertLegacyConfigExpression(config, migration) {
...
+ * @param {string[]} options.ignorePatterns An array of glob patterns to ignore.
+ * @param {boolean} options.gitignore `true` to include contents of a .gitignore file.
+ * @param {"9"|"10"} options.targetVersion The target version of ESLint for the migration.
  * @returns {{code:string,messages:Array<string>,imports:Map<string,MigrationImport>}} The migrated config and
  * any messages to display to the user.
  */
 export function migrateConfig(
 	config,
-	{ sourceType = "module", ignorePatterns, gitignore = false } = {},
+	{ sourceType, ignorePatterns, gitignore, targetVersion },
```
lumirlumir, line 1456: I think if we want to change the internal API design for conciseness, we should explicitly add an `"exports": {}` field in `package.json` for `@eslint/migrate-config` to fully prevent subpath imports by end users.

For example, since there is currently no `exports` field in `package.json`, users can import internal modules via subpaths like this:

```js
import eslintConfigReactApp from "@eslint/migrate-config/src/compat-configs.js";

console.log(eslintConfigReactApp);
```

If we run the code above, the output is currently `[ 'eslint-config-react-app' ]`, which means subpath imports are accessible to end users.

However, if we add `"exports": {}` to `package.json`, like this:

```diff
  "bin": {
    "eslint-migrate-config": "./src/migrate-config-cli.js"
  },
+ "exports": {},
  "files": [
    "src"
  ],
```

users can no longer access subpaths, and it throws `ERR_PACKAGE_PATH_NOT_EXPORTED` instead. So this change would help ensure internal API stability by preventing access through undocumented paths.

(Since this PR is a breaking change, I believe this adjustment would be acceptable.)

packages/migrate-config/src/migrate-config.js
```diff
@@ -395,15 +411,34 @@ function convertGlobPattern(pattern) {
...
+				.get("@eslint/compat")
+				.bindings.push("includeIgnoreFile");
+		}
+
+		code = `includeIgnoreFile(gitignorePath)`;
 	} else {
-		migration.imports
-			.get("@eslint/compat")
-			.bindings.push("includeIgnoreFile");
+		if (!migration.imports.has("eslint/config")) {
```
lumirlumir, line 431: Question:

Should we consider the case where `@eslint/config-helpers` is already being used?

For example, this repo currently imports `includeIgnoreFile` from `@eslint/config-helpers`:

https://github.com/eslint/rewrite/blob/33fc0b32d9ab0965dfe1d2b29d3935057ba8d2ca/eslint.config.js#L2

With the current implementation, if a config already imports helpers from `@eslint/config-helpers` and does not import from `eslint/config`, the migration would add a new `eslint/config` import path instead of reusing the existing helper import source.

I'm not strongly suggesting we need to handle this, since the generated config already seems to standardize on `eslint/config`. I just wanted to raise it as a possible edge case and get another opinion.

lumirlumir (changes requested):
Thanks for getting started on this! I’ve left a few comments.

packages/migrate-config/src/migrate-config.js
kirkwaiblinger: My 2c would be that the generated code here should definitely standardize on `eslint/config`, and not expose `@eslint/config-helpers` to users.

I'd take this repo as an anomaly, since it defines `@eslint/config-helpers` itself (and is not part of a monorepo with `eslint`). As a consequence, importing from `@eslint/config-helpers` is required in order for updates to those APIs to immediately be reflected within this repo itself for dogfooding. This situation is unique to the eslint/rewrite repo.

(note the internal import of config-helpers)
https://github.com/eslint/rewrite/blob/0a4fef6f5a2ee9f53facef5f4b3f4d1bef0b8e2f/package.json#L43

packages/migrate-config/src/migrate-config.js
kirkwaiblinger: Makes sense, added `exports: {}`.

packages/migrate-config/src/migrate-config.js
mdjermanovic: > Should we consider the case where `@eslint/config-helpers` is already being used?

`@eslint/migrate-config` converts legacy eslintrc configs to new eslint.config.js configs. I believe `@eslint/config-helpers` isn't applicable in eslintrc configs, so it doesn't seem it can happen that `@eslint/config-helpers` is already used?

mdjermanovic (approved):
LGTM, thanks! Leaving open for @lumirlumir to verify.

lumirlumir (approved):
LGTM, thanks!

PR #498, opened 2026-08-18
perf: speed up ConfigArray config resolution

nzakas:
## What

Speeds up `getConfigWithStatus()` and `isDirectoryIgnored()` in `@eslint/config-array` by roughly **4x on posix paths and 1.7x on Windows paths**, and adds a benchmark script for measuring config resolution performance.

Benchmark (1,500 unique file paths against an 8-config array, 100 runs, same machine, mean time per run):

| Scenario | Before | After | Improvement |
|---|---|---|---|
| posix paths | 13.78ms | 3.53ms | 74% (3.9x) |
| Windows paths | 16.56ms | 9.65ms | 42% (1.7x) |

## Changes

- **Compiled regex matching.** `doMatch()` now compiles each pattern once via `Minimatch#makeRe()` and evaluates `regexp.test()` instead of `Minimatch#match()`, which re-splits the file path and walks the pattern segment-by-segment on every call. Three semantic differences between `match()` and the compiled regex are compensated exactly:
  - negation is applied outside the regex (compiled from the `!`-stripped pattern), so normal and `flipNegate` matching share one code path;
  - a path with a trailing slash retries without it (`"foo/"` matches pattern `"foo"`);
  - for trailing-globstar patterns, the regex's optional final group is made required (`"a/**"` must not match `"a"`), with a fallback to `match()` for patterns that can't be rewritten safely (e.g. brace expansions with a trailing globstar).
- **Fast posix relative paths.** `toRelativePath()` short-circuits for already-normalized absolute posix paths: a `slice()` when the path is under the base, or a ported allocation-light `relative()` walk when it isn't. Windows paths keep the original code path.
- **Per-config metadata cache.** The universal/non-universal `files` partition and the global-ignores key check are computed once per config object (WeakMap) instead of per file x per config.
- **Allocation-free hot loops.** The `reduce()`/`some()`/`filter()` callbacks, per-call regex literal, and temporary `{ files, ignores }` wrapper objects in `shouldIgnorePath()`/`pathMatches()` are replaced with plain indexed loops (`matchesIgnores()`, `pathMatchesFiles()`).
- **Debug guards.** Template-literal debug messages are only built when `debug.enabled` is true.
- **Benchmark.** `npm run bench` in `packages/config-array` (options: `--windows`, `--runs=N`); see `benchmarks/README.md`.

## Verification

- All 243 existing unit tests pass, along with the full build, type tests, lint, and Prettier.
- Pattern-matching fuzz vs. raw minimatch across 5,700 pattern/path combinations: the only divergences involve `..` segments, which are normalized away or rejected as external before matching.
- Relative-path fuzz vs. `@jsr/std__path` across 200k random path pairs: zero mismatches.
- Differential fuzz of the full public API (old vs. new implementation, 80k randomized trials including `basePath` configs, negated patterns, brace patterns, and function matchers, on both posix and Windows base paths): zero mismatches in status, merged config, or directory-ignore results.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/config-array/benchmarks/config-resolution.bench.js
+++ b/packages/config-array/benchmarks/config-resolution.bench.js
@@ -28,9 +28,27 @@ import { ConfigArray } from "../src/index.js";
 // Options
 //------------------------------------------------------------------------------
 
+/**
+ * Parses the value of the `--runs` option.
+ * @param {string} value The raw option value.
+ * @returns {number} The number of runs to perform.
+ * @throws {Error} When the value is not a positive integer.
+ */
+function parseRuns(value) {
+	const runs = Number(value);
+
+	if (!Number.isInteger(runs) || runs < 1) {
+		throw new Error(
+			`Invalid --runs value ${JSON.stringify(value)}: expected a positive integer.`,
+		);
+	}
+
+	return runs;
+}
+
 const useWindowsPaths = process.argv.includes("--windows");
 const runsArg = process.argv.find(arg => arg.startsWith("--runs="));
-const RUNS = runsArg ? Number(runsArg.slice("--runs=".length)) : 100;
+const RUNS = runsArg ? parseRuns(runsArg.slice("--runs=".length)) : 100;
 const WARMUP_RUNS = 10;
 const FILE_COUNT = 1500;
 const DIRECTORY_CHECK_COUNT = 50;
--- a/packages/config-array/benchmarks/config-resolution.bench.js
+++ b/packages/config-array/benchmarks/config-resolution.bench.js
@@ -1,0 +1,264 @@
+/**
+ * @fileoverview Benchmark for ConfigArray config resolution.
+ *
+ * Measures the cost of `getConfigWithStatus()` and `isDirectoryIgnored()`
+ * across many unique file paths against a realistic flat-config-style
+ * config array. Each run uses a fresh `ConfigArray` instance so that the
+ * per-path result cache doesn't hide resolution cost; a second pass over
+ * the same paths exercises the cache-hit path.
+ *
+ * Usage:
+ *   node benchmarks/config-resolution.bench.js [--windows] [--runs=N]
+ *
+ * Options:
+ *   --windows   Use Windows-style paths (drive letter and backslashes).
+ *   --runs=N    Number of timed runs (default 100).
+ *
+ * @author Nicholas C. Zakas
+ */
+
+//------------------------------------------------------------------------------
+// Imports
+//------------------------------------------------------------------------------
+
+import { performance } from "node:perf_hooks";
+import { ConfigArray } from "../src/index.js";
+
+//------------------------------------------------------------------------------
+// Options
+//------------------------------------------------------------------------------
+
+const useWindowsPaths = process.argv.includes("--windows");
+const runsArg = process.argv.find(arg => arg.startsWith("--runs="));
+const RUNS = runsArg ? Number(runsArg.slice("--runs=".length)) : 100;
+const WARMUP_RUNS = 10;
+const FILE_COUNT = 1500;
+const DIRECTORY_CHECK_COUNT = 50;
+
+const BASE_PATH = useWindowsPaths ? "C:\\project" : "/project";
+const SEPARATOR = useWindowsPaths ? "\\" : "/";
+
+//------------------------------------------------------------------------------
+// Fixtures
+//------------------------------------------------------------------------------
+
+/**
+ * A schema for the custom config keys used by the fixture configs.
+ * @type {Object}
+ */
+const customSchema = {
+	languageOptions: {
+		required: false,
+		merge(a = {}, b = {}) {
+			return { ...a, ...b };
+		},
+		validate() {},
+	},
+	rules: {
+		required: false,
+		merge(a = {}, b = {}) {
+			return { ...a, ...b };
+		},
+		validate() {},
+	},
+};
+
+/**
+ * Creates a realistic flat-config-style array of config objects.
+ * @returns {Array<Object>} The config objects.
+ */
+function makeConfigs() {
+	return [
+		// global ignores
+		{
+			ignores: [
+				"**/node_modules/",
+				"dist/**",
+				"coverage/**",
+				"**/*.min.js",
+				"!dist/keep.js",
+			],
+		},
+
+		// universal config
+		{
+			name: "base",
+			languageOptions: { ecmaVersion: 2026 },
+		},
+
+		// all JS files
+		{
+			name: "js",
+			files: ["**/*.js", "**/*.mjs", "**/*.cjs"],
+			ignores: ["**/*.config.js"],
+			rules: { semi: 2, quotes: 1 },
+		},
+
+		// TS files
+		{
+			name: "ts",
+			files: ["src/**/*.ts", "src/**/*.tsx"],
+			rules: { "no-var": 2 },
+		},
+
+		// tests, including an AND pattern
+		{
+			name: "tests",
+			files: ["tests/**/*.js", ["**/*.test.js", "tests/**"]],
+			rules: { "prefer-const": 1 },
+		},
+
+		// universal pattern config (triggers universal files handling)
+		{
+			name: "universal-src",
+			files: ["src/**"],
+			rules: { "no-console": 1 },
+		},
+
+		// config with a base path
+		{
+			name: "based",
+			basePath: "packages/lib",
+			files: ["**/*.js"],
+			rules: { "no-alert": 2 },
+		},
+
+		// markdown files
+		{
+			name: "docs",
+			files: ["docs/**/*.md"],
+			rules: {},
+		},
+	];
+}
+
+/**
+ * Creates a list of mostly-unique file paths that are a mix of matched,
+ * ignored, and unconfigured files.
+ * @returns {Array<string>} The file paths.
+ */
+function makeFilePaths() {
+	const dirs = [
+		"src",
+		"src/components",
+		"src/utils/deep/nested",
+		"tests",
+		"tests/unit",
+		"lib",
+		"docs",
+		"packages/lib/src",
+		"node_modules/foo",
+		"dist/assets",
+		"coverage/lcov",
+	];
+	const exts = [
+		".js",
+		".mjs",
+		".ts",
+		".tsx",
+		".md",
+		".txt",
+		".test.js",
+		".min.js",
+		".config.js",
+	];
+	const paths = [];
+
+	for (let i = 0; i < FILE_COUNT; i++) {
+		const dir = dirs[i % dirs.length].replaceAll("/", SEPARATOR);
+		const ext = exts[i % exts.length];
+
+		paths.push(`${BASE_PATH}${SEPARATOR}${dir}${SEPARATOR}file${i}${ext}`);
+	}
+
+	return paths;
+}
+
+const filePaths = makeFilePaths();
+
+//------------------------------------------------------------------------------
+// Benchmark
+//------------------------------------------------------------------------------
+
+/**
+ * Runs one full resolution pass over all file paths with a fresh
+ * `ConfigArray` instance.
+ * @returns {Promise<{matched: number, ignored: number, other: number}>} Counts
+ *		of file statuses, used as a sanity check.
+ */
+async function once() {
+	const configs = new ConfigArray(makeConfigs(), {
+		basePath: BASE_PATH,
+		schema: customSchema,
+		extraConfigTypes: ["array", "function"],
+	});
+
+	await configs.normalize();
+
+	let matched = 0,
+		ignored = 0,
+		other = 0;
+
+	// first pass: every lookup is a cache miss
+	for (const filePath of filePaths) {
+		const { status } = configs.getConfigWithStatus(filePath);
+
+		if (status === "matched") {
+			matched++;
+		} else if (status === "ignored") {
+			ignored++;
+		} else {
+			other++;
+		}
+	}
+
+	// second pass: every lookup is a cache hit
+	for (const filePath of filePaths) {
+		configs.getConfigWithStatus(filePath);
+	}
+
+	// directory ignore checks
+	for (let i = 0; i < DIRECTORY_CHECK_COUNT; i++) {
+		configs.isDirectoryIgnored(
+			`${BASE_PATH}${SEPARATOR}src${SEPARATOR}dir${i}`,
+		);
+		configs.isDirectoryIgnored(
+			`${BASE_PATH}${SEPARATOR}node_modules${SEPARATOR}pkg${i}`,
+		);
+	}
+
+	return { matched, ignored, other };
+}
+
+for (let i = 0; i < WARMUP_RUNS; i++) {
+	await once();
+}
+
+if (globalThis.gc) {
+	globalThis.gc();
+}
+
+const times = [];
+let sanity;
+
+for (let i = 0; i < RUNS; i++) {
+	const start = performance.now();
+
+	sanity = await once();
+	times.push(performance.now() - start);
+}
+
+times.sort((a, b) => a - b);
+
+const mean = times.reduce((a, b) => a + b, 0) / times.length;
+const median = times[Math.floor(times.length / 2)];
+const p10 = times[Math.floor(times.length * 0.1)];
+
+/* eslint-disable no-console -- CLI benchmark reports results to stdout */
+console.log(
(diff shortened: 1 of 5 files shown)


packages/config-array/src/config-array.js
```diff
@@ -221,31 +251,116 @@ function assertValidBaseConfig(config, index) {
...
+				const compiled = rawMatcher.makeRe();
+
+				if (compiled && compiled.source.endsWith(")?$")) {
+					regexp = new RegExp(
+						`${compiled.source.slice(0, -2)}$`,
+						compiled.flags,
+					);
+				}
+			}
+		}
```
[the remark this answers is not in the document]
nzakas: Good catch, and investigating it turned up something worse than the recursion bound: `makeRe()` is **lossy** for these patterns, not just unbounded.

For a pattern with more than one globstar after the first path segment, `makeRe()` silently drops all but the first one. `src/**/test/**/*.js` compiled to a regex equivalent to `src/**/test/*.js`, so `src/test/x/a.js` stopped matching — a silent behavior regression, not just a performance/safety question. My original differential fuzzing missed it because the pattern corpus never had two internal globstars.

Fixed in 232818fb9 by gating the fast path on `canCompileToRegExp()`: patterns with more than one non-leading globstar stay on `match()`, which resolves your `maxGlobstarRecursion` point as well, since those are exactly the patterns the bound applies to. Patterns with a leading globstar plus at most one later globstar compile correctly and still take the fast path, so `**/*.js`, `src/**`, `**/node_modules/**`, and `packages/*/src/**` are unaffected and the speedup is unchanged (posix 15.4ms → 3.3ms median, interleaved A/B).

Also added regression tests for multi-globstar patterns; 3 of the 4 fail without the gate. Verified with an exhaustive differential run against the pre-change implementation: 1,554 generated patterns × 680 paths = 1.06M comparisons, zero mismatches.

On backtracking for the patterns that still compile: I measured `match()` vs `RegExp#test()` on 40KB adversarial paths (20k segments). The regex stays flat at well under 1ms, while `match()` degrades to ~250-280ms for `**/node_modules/**` and `**/a/**`. So for the shapes remaining on the fast path, the regex is the better-behaved of the two on hostile input.

packages/config-array/benchmarks/config-resolution.bench.js
```diff
@@ -0,0 +1,264 @@
...
+import { performance } from "node:perf_hooks";
+import { ConfigArray } from "../src/index.js";
+
+//------------------------------------------------------------------------------
+// Options
+//------------------------------------------------------------------------------
+
+const useWindowsPaths = process.argv.includes("--windows");
+const runsArg = process.argv.find(arg => arg.startsWith("--runs="));
+const RUNS = runsArg ? Number(runsArg.slice("--runs=".length)) : 100;
```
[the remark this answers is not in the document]
nzakas: Fixed in 232818fb9. `--runs` now goes through a `parseRuns()` helper that rejects anything that isn't a positive integer:

```
$ node benchmarks/config-resolution.bench.js --runs=0
Error: Invalid --runs value "0": expected a positive integer.
```

Same for `abc`, `2.5`, and `-5`, so the fractional-value mismatch you noted is covered too.

nzakas:
@mdjermanovic @fasttime

packages/config-array/src/config-array.js
```diff
@@ -483,6 +636,140 @@ function normalizeSync(
...
+ * `@jsr/std__path/posix` without the redundant `resolve()` calls, for use
+ * when both paths are known to be normalized absolute paths.
+ * @param {string} from The resolved path to start from.
+ * @param {string} to The resolved path to reach.
+ * @returns {string} The relative path.
+ */
+function posixRelativeResolved(from, to) {
+	if (from === to) {
+		return "";
+	}
```
fasttime, line 654: I think this case can never happen because it's already handled by the caller:

https://github.com/eslint/rewrite/blob/232818fb9fe7bb9e1f7eaecbca569f0775f1d5b1/packages/config-array/src/config-array.js#L755-L764

packages/config-array/src/config-array.js
```diff
@@ -538,95 +893,211 @@ function shouldIgnorePath(
...
+ * Retrieves cached metadata for a config object, calculating it first
+ * if not already cached.
+ * @param {Object} config The config object to get metadata for.
+ * @returns {ConfigMetadata} The metadata for the config object.
+ */
+function getConfigMetadata(config) {
+	// non-object configs cannot be used as WeakMap keys
+	if (typeof config !== "object" || config === null) {
+		return calculateConfigMetadata(Object(config));
+	}
```
fasttime, line 1089: Since `getConfigMetadata` is only called after `assertNormalized`, I think `config` will be always a non-null object.

packages/config-array/tests/config-array.test.js
fasttime: Based on `npm run test:coverage`, coverage for this package has dropped significantly.

On the `main` branch:

```text
Statements   : 98.62% ( 1510/1531 )
Branches     : 97.64% ( 249/255 )
Functions    : 93.33% ( 42/45 )
Lines        : 98.62% ( 1510/1531 )
```

On this PR branch rebased on `main`:

```text
Statements   : 95.66% ( 1940/2028 )
Branches     : 91.95% ( 320/348 )
Functions    : 94.11% ( 48/51 )
Lines        : 95.66% ( 1940/2028 )
```

packages/config-array/src/config-array.js
```diff
@@ -1030,7 +1499,9 @@ export class ConfigArray extends Array {
 		);
 
 		if (EXTERNAL_PATH_REGEX.test(relativeToBaseFilePath)) {
-			debug(`No config for file ${filePath} outside of base path`);
+			if (debug.enabled) {
+				debug(`No config for file ${filePath} outside of base path`);
+			}
```
fasttime, line 1504:
```suggestion
			debug("No config for file %s outside of base path", filePath);
```

Instead of checking `debug.enabled` to avoid the string interpolation when debugging is disabled, I'd suggest calling `debug` with a placeholder in the format string. The format string is only processed when debugging is enabled. This change also avoids incorrect output when `filePath` contains certain special reserved substrings like `%%`, `%d`, etc. (see https://github.com/debug-js/debug#formatters), which `debug` treats as formatters when they appear in the first argument.

packages/config-array/src/config-array.js
```diff
@@ -1041,7 +1512,9 @@ export class ConfigArray extends Array {
 
 		// check if this should be ignored due to its directory
 		if (this.isDirectoryIgnored(this.#path.dirname(filePath))) {
-			debug(`Ignoring ${filePath} based on directory pattern`);
+			if (debug.enabled) {
+				debug(`Ignoring ${filePath} based on directory pattern`);
+			}
```
fasttime, line 1517:
```suggestion
			debug("Ignoring %s based on directory pattern", filePath);
```
Here, too, and for the other usages below, it's better to use a format string. This would avoid:
- An unnecessary interpolation when debugging is not enabled
- Incorrect output in edge cases
- Some uncovered branches and statements (see the next comment)

packages/config-array/src/config-array.js
```diff
@@ -1054,7 +1527,9 @@ export class ConfigArray extends Array {
 				path: this.#path,
 			})
 		) {
-			debug(`Ignoring ${filePath} based on file pattern`);
+			if (debug.enabled) {
+				debug(`Ignoring ${filePath} based on file pattern`);
+			}
```
fasttime:
```suggestion
			debug("Ignoring %s based on file pattern", filePath);
```

Similar to the previous suggestion.

packages/config-array/src/config-array.js
```diff
@@ -1065,146 +1540,162 @@ export class ConfigArray extends Array {
...
+					this.#path,
 				);
-				return;
+
+				if (EXTERNAL_PATH_REGEX.test(relativeFilePath)) {
+					if (debugEnabled) {
+						debug(
+							`Skipped config found for ${filePath} (based on config's base path: ${config.basePath}`,
+						);
+					}
```
fasttime:
```suggestion
					debug(
						"Skipped config found for %s (based on config's base path: %s",
						filePath,
						config.basePath,
					);
```

Here, too, and for the usages cases below, it's better to use a format string both to avoid an unnecessary interpolation and also to avoid incorrect output in some edge cases.

packages/config-array/tests/config-array.test.js
fasttime, line 1: Based on `npm run test:coverage`, code coverage for this PR seems significantly lower than on `main`.

On the `main` branch:

```text
Statements   : 98.62% ( 1510/1531 )
Branches     : 97.64% ( 249/255 )
Functions    : 93.33% ( 42/45 )
Lines        : 98.62% ( 1510/1531 )
```

On this PR branch rebased on `main`:

```text
Statements   : 95.66% ( 1940/2028 )
Branches     : 91.95% ( 320/348 )
Functions    : 94.11% ( 48/51 )
Lines        : 95.66% ( 1940/2028 )
```

Those are the results I'm seeing locally on MacOS with Node.js 24 (I haven't tested this on Windows yet). Since this PR adds a fair amount of new code, some drop seems normal, but it would be good to add unit tests to make sure we're not missing coverage on branches that don't actually occur in practice.

packages/config-array/src/config-array.js
nzakas: Agreed, thanks. Looking at it more closely, several other branches in that function were unreachable for the same reason: it was a full port of `std/path`'s `relative()`, but the caller has already handled equal paths, the root base path, and paths inside the base path.

In ea804d5cc I replaced it with a smaller `posixRelativeOutside()` that only handles what it can actually receive (a path outside the base path), so the result always starts with `..`. Every branch in it is now covered. I verified it against `std/path`'s `relative()` on ~550k generated path pairs targeted at the edge cases (root, ancestor, equal, inside, partial-segment siblings), with no mismatches.

I also removed the `basePath.charCodeAt(0) !== SLASH_CHAR_CODE` check in `fastPosixRelative()`, which was unreachable for a similar reason: posix mode is only selected for base paths starting with `/`, and config base paths are either absolute or resolved.

packages/config-array/src/config-array.js
nzakas: Agreed, removed in ea804d5cc. One small caveat I ran into while checking: `assertNormalized()` doesn't strictly imply validation, because `new ConfigArray(configs, { normalized: true })` skips `assertValidBaseConfig()`.

So rather than rely on that, I made the metadata lookup lazy. It's now only called in branches that already require a truthy `config.files` or `config.ignores`, which rules out primitives structurally. A `null` config would already throw on `config.basePath` a few lines earlier, the same as on `main`. It also skips the WeakMap lookup for universal configs.

packages/config-array/src/config-array.js
nzakas: Applied this and your other suggestion, and converted the remaining calls in ea804d5cc, including the pre-existing `No matching configs found` call, which had the same `%` issue. That removed all the `debug.enabled` guards.

I checked with `DEBUG=@eslint/config-array` that a path like `/p/100%d.js` now prints unchanged. Converting also fixed a missing closing parenthesis in the `based on config's base path` message.

packages/config-array/tests/config-array.test.js
nzakas: Thanks for checking. You were right that the drop came from branches that don't happen in practice, plus a platform effect: the suite's `basePath` comes from `import.meta.url`, so on macOS the Windows branches of the new code were never run, and on Windows the posix fast path wasn't.

Changes in ea804d5cc:
- Removed the unreachable branches (the `debug.enabled` guards, the `getConfigMetadata()` guard, and the unused parts of the relative-path function; see the other threads).
- Added tests with explicit posix and Windows base paths, covering config `basePath` matching, global ignores with `basePath`, ancestor directories, a root base path, trailing slashes, paths needing full resolution, and comment patterns. The platform-specific functions are fully covered by these tests on their own, so the result shouldn't depend on the OS. All the new tests also pass against `main`'s implementation, so they pin existing behavior.

`npm run test:coverage` after merging the latest `main` (Windows, Node 22):

```text
            main     PR before   PR now
Statements  98.62%   95.66%      98.93%
Branches    97.64%   91.95%      98.21%
Functions   93.33%   94.11%      94.11%
Lines       98.62%   95.66%      98.93%
```

The remaining uncovered lines are all ones that are also uncovered on `main`.

packages/config-array/src/config-array.js
```diff
@@ -483,6 +636,114 @@ function normalizeSync(
 	return configs;
 }
 
+// Detects repeated slashes or `.`/`..` segments in an absolute posix path.
+const POSIX_UNSAFE_SEGMENT_REGEX = /\/\/|\/\.{1,2}(?:\/|$)/u;
```
fasttime: We could move this constant up before the first class declaration (`ConfigError`) for consistency with the style used in this file.

packages/config-array/src/config-array.js
```diff
@@ -538,95 +867,206 @@ function shouldIgnorePath(
...
 	}
 
 	return filePathMatchesPattern;
 }
 
+/*
+ * Pattern to detect universal file patterns: `*`, patterns starting with
+ * `!`, or patterns ending with `/*` or `/**`.
+ */
+const UNIVERSAL_PATTERN_REGEX = /^\*$|^!|\/\*{1,2}$/u;
```
fasttime: This should be also moved up where the other named constants are defined in this file.

packages/config-array/src/config-array.js
```diff
@@ -538,95 +867,206 @@ function shouldIgnorePath(
...
+ * 		in `files`, or `null` if the config has no `files`.
+ * @property {FilesMatcher[]|null} nonUniversalFiles The non-universal patterns
+ * 		found in `files`, or `null` if the config has no `files`.
+ */
+
+/**
+ * A cache for config object metadata.
+ * @type {WeakMap<Object, ConfigMetadata>}
+ */
+const configMetadataCache = new WeakMap();
```
fasttime: This one, too.

packages/config-array/src/config-array.js
```diff
@@ -44,14 +44,28 @@ import { filesAndIgnoresSchema } from "./files-and-ignores-schema.js";
...
+ * to re-split the file path the way `Minimatch#match()` does. When a pattern
+ * cannot be compiled to a regular expression (comments, empty patterns,
+ * syntax errors), `regexp` is `null` and the `Minimatch` instance is used
+ * directly.
+ * @typedef {Object} MatcherEntry
+ * @property {Minimatch} matcher The minimatch instance for the pattern.
+ * @property {RegExp|null} regexp The compiled regular expression, if available.
+ * @property {boolean} negate True if the regular expression's raw result
+ * 		should be negated to produce the match result.
+ */
```
fasttime: This typedef should be moved up in the "types" section.

packages/config-array/src/config-array.js
```diff
@@ -538,95 +867,206 @@ function shouldIgnorePath(
...
+ * on every file path lookup, so it's computed once per config object and
+ * cached.
+ * @typedef {Object} ConfigMetadata
+ * @property {boolean} isGlobalIgnores True if the config object only has
+ * 		`ignores` (aside from meta fields) and therefore acts as global ignores.
+ * @property {FilesMatcher[]|null} universalFiles The universal patterns found
+ * 		in `files`, or `null` if the config has no `files`.
+ * @property {FilesMatcher[]|null} nonUniversalFiles The non-universal patterns
+ * 		found in `files`, or `null` if the config has no `files`.
+ */
```
fasttime: This typedef should be moved up in the `types` section.

packages/config-array/src/config-array.js
```diff
@@ -221,31 +251,154 @@ function assertValidBaseConfig(config, index) {
...
 	}
 
-	return matcher.match(filepath);
+	const { regexp } = entry;
+
+	/*
+	 * The empty string requires `Minimatch#match()`'s segment-based
+	 * handling, so defer to it in that case.
+	 */
+	if (regexp !== null && filepath !== "") {
```
fasttime: Not sure if `filepath` can be an empty string at this point. It should be always a relative path, hence not empty.

fasttime: Not sure if `filepath` can be an empty string at this point since it's supposed to be a relative path. Maybe the check could be removed:


```suggestion
	if (regexp !== null) {
```

fasttime (commented):
When I run the benchmark on my machine I can see an approximately 75% speedup for posix paths and approximately 35% for Windows paths, which is broadly consistent with the figures reported in the PR description.

I left a few more comments for the coding agent to follow the stylistic conventions in this repo, and one suggestion to remove a possibly constant condition.

PR #502, opened 2026-08-26
ci: use hoisted linker for bun installs

sethamus:
#### Prerequisites checklist

#### AI acknowledgment

This PR fixes the `ci-bun` job, which has been failing on every branch since Bun 1.4.0 was released.

Bun 1.4.0 made the isolated linker the default for monorepos whose lockfile has `configVersion: 1`. We don't commit `bun.lock`, so CI always generates a fresh one and always gets isolated linking.

Added a root `bunfig.toml` pinning Bun to the hoisted linker.

#### Related Issues

https://github.com/eslint/rewrite/actions/runs/32868554819/job/97869747114

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -1,0 +1,2 @@
+[install]
+linker = "hoisted"


fasttime:
Thanks for the PR! I think this change makes sense as a quick fix for CI. We should eventually revisit the workflow, since it doesn't seem especially useful to check whether the repo can be built from source with Bun. A more useful check, which is currently missing, would verify that the built packages can be installed as dependencies in Bun with the default settings.

fasttime (approved):
LGTM, thanks!

PR #503, opened 2026-08-27
fix: preserve markVariableAsUsed return value in compat context

OskarEichler:
## Fix

Return the boolean from `SourceCode#markVariableAsUsed()` through the legacy compatibility context. The wrapper currently marks the variable but always returns `undefined`, so legacy rules cannot distinguish a successful lookup from a missing variable.

This is a one-line change. Scope selection and the marking side effect remain unchanged. No public API removals, dependency changes, or breaking changes are introduced.

## Verification

- All 332 existing `@eslint/compat` tests pass.
- Repository build, targeted source lint, and repository type checks pass.
- An inline real-ESLint-10 reproduction checks existing and missing names in global and nested scopes: upstream returns `undefined`; this change returns `true` / `false` as expected. Existing variables remain marked as used.
- `git diff --check` passes. No test files were added or changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the success status returned when marking variables as used.
  * Correctly reports whether an existing or missing variable was marked.

* **Tests**
  * Added regression coverage for variable usage results.

### DIFF
--- a/packages/compat/src/fixup-rules.js
+++ b/packages/compat/src/fixup-rules.js
@@ -295,7 +295,10 @@ export function fixupRule(ruleDefinition) {
 			},
 
 			markVariableAsUsed(variable) {
-				return compatSourceCode.markVariableAsUsed(variable, currentNode);
+				return compatSourceCode.markVariableAsUsed(
+					variable,
+					currentNode,
+				);
 			},
 		});
 
--- a/packages/compat/src/fixup-rules.js
+++ b/packages/compat/src/fixup-rules.js
@@ -295,7 +295,7 @@ export function fixupRule(ruleDefinition) {
 			},
 
 			markVariableAsUsed(variable) {
-				compatSourceCode.markVariableAsUsed(variable, currentNode);
+				return compatSourceCode.markVariableAsUsed(variable, currentNode);
 			},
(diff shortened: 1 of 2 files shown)


DMartens:
Thank you for the PR.
Can you please also add a simple test case for this?

DMartens (dismissed):
Changes LGTM, thanks. Leaving open in case others want to review it.

packages/compat/src/fixup-rules.js
```diff
@@ -295,7 +295,7 @@ export function fixupRule(ruleDefinition) {
 			},
 
 			markVariableAsUsed(variable) {
-				compatSourceCode.markVariableAsUsed(variable, currentNode);
+				return compatSourceCode.markVariableAsUsed(variable, currentNode);
```
mdjermanovic:
```suggestion
				return compatSourceCode.markVariableAsUsed(
					variable,
					currentNode,
				);
```

To fix the formatting error.

OskarEichler:
Applied the requested formatting to the forwarded markVariableAsUsed call. The branch includes the return-value regression test and the current check suite is green.

DMartens (approved):

PR #505, opened 2026-08-31
docs: clarify parameter description in `getText`

lumirlumir:
#### Prerequisites checklist

#### AI acknowledgment

This is a small documentation clarification.

While working on my Markdown plugin, I found that unlike `TextSourceCodeBase#getRange` and `TextSourceCodeBase#getLoc`, `TextSourceCodeBase#getText` does not throw an error when `undefined` is passed as its argument, and instead returns the full text.

I think we could clarify this behavior in the JSDoc so users won’t be confused by the difference in behavior.

I simply updated the JSDoc to clarify the parameter description.

#### Related Issues

N/A

N/A

## Summary by CodeRabbit

* **Documentation**
  * Clarified that calling `getText` without a node, or with any falsy node value, returns the complete source text.


packages/plugin-kit/src/source-code.js
```diff
@@ -605,7 +605,7 @@ export class TextSourceCodeBase {
 
 	/**
 	 * Gets the source code for the given node.
-	 * @param {Options['SyntaxElementWithLoc']} [node] The AST node to get the text for.
+	 * @param {Options['SyntaxElementWithLoc']} [node] The AST node to get the text for. If `undefined` or omitted, the entire source text is returned.
```
DMartens: `undefined` is incorrect as the method checks whether `node` is "truthy".
So the whole text is also returned for `null` or an empty string.
Side note: We have the same description in th JavaScript source code and the description should also be updated there

lumirlumir: Updated it in https://github.com/eslint/rewrite/pull/505/changes/a62fd604fa3e39843397ebed47e10a9071e37f72.

DMartens (approved):
Changes LGTM, thanks.

